PR #130, opened 2019-12-13
Update default http port

### LINKED ISSUE #129: Gasket start/local will not start
## 🐛 Bug Report
running `npx gasket local` fails to run http/https servers. This is due to it attempting to run on port 80 which is a privileged port. 

## To Reproduce
`npm i -g @gasket/cli`
`gasket create *app-name* --presets @gasket/nextjs`
`npx gasket local`

## Expected Behavior
expects to work out of the box on a basic setup/install with out needing to modify `http/https`

## Environment
Manjaro Linux (Arch based OS) local environment

rxmarbles:
## Summary
Fixes #129, updates `configOpts.http` to port 8080 if neither exist in the gasket config

## Changelog
 
- Update default http port if not defined to 8080

## Test Plan

updated unit test to reflect change

### DIFF
--- a/packages/gasket-plugin-https/index.js
+++ b/packages/gasket-plugin-https/index.js
@@ -34,7 +34,8 @@ async function start(gasket) {
   // create-servers does not support http or https being `null`
   if (http) configOpts.http = http;
   if (https) configOpts.https = https;
-  if (!http && !https) configOpts.http = 80;
+  // Default port to non-essential port on creation
+  if (!http && !https) configOpts.http = 8080;
 
   const serverOpts = await gasket.execWaterfall('createServers', configOpts);
   const { healthcheck, ...terminusDefaults } = await gasket.execWaterfall('terminus', {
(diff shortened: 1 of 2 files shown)


rxmarbles:
@SivanMehta that's a good point we could do that as well, however port 80 is still privileged and should not be the default port unless its in PROD perhaps adding more logic around it vs changing the port would make sense there then?

SivanMehta:
@rxmarbles perhaps the correct answer is _both_.

- `npm run local`: `:80` should not be the defeault because it's privileged on (most) development machines. 
- `npm start`: This command is aimed at production-facing uses, so `:80` is a more appropriate default here.

kinetifex:
@rxmarbles I think @SivanMehta has a good suggestion here. Leave 80 as the default for production cases and using start command, with the local env pre-configured in gasket.config to be 8080. This is probably the best _it just works_ setup for handling both local and production deployments.

rxmarbles:
@SivanMehta @kinetifex updates have been pushed ready for another review 😄

PR #144, opened 2020-02-06
Fix missing await

kinetifex:
## Summary

There was a strange race condition when using `gasket local` in an app with both intl and service-worker plugins. Tracked it down to this missing await.

## Changelog

- Fixed missing await when saving `locales-manifest.json` file

## Test Plan

- Added unit test. Integration test in repro app with `npm link`.


packages/gasket-plugin-intl/lib/builder.spec.js
```diff
@@ -203,13 +203,19 @@ describe('Builder', function () {
...
+        fsUtils.saveJsonFile = jest.fn().mockReturnValue(Promise.resolve());
         builder.processFiles = jest.fn();
       });
 
       it('calls processFiles for each directory', async function () {
         await builder.processDirs(buildDirs);
         expect(builder.processFiles.mock.calls).toHaveLength(2);
       });
+
+      it('saves locales manifest file', async function () {
```
msluther: You could also add a test to verify that you get the promise rejection when `fsUtils.saveJsonFile` results in a rejection.

PR #153, opened 2020-02-26
Add more accurate desc of redux client-side of config

blackswanny:
## Summary

There is some difference between example of client-side redux reducer injection of config env and same for gasket non-run time and server-side example. I fixed it

### DIFF
--- a/packages/gasket-plugin-config/README.md
+++ b/packages/gasket-plugin-config/README.md
@@ -149,8 +149,8 @@ module.exports = {
 
 If you need access to config values in client-side code, this can be done
 through your redux store. The config plugin looks for a `redux` property of your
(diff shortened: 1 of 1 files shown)


packages/gasket-plugin-config/README.md
```diff
@@ -149,8 +149,25 @@ module.exports = {
 
 If you need access to config values in client-side code, this can be done
 through your redux store. The config plugin looks for a `redux` property of your
-configuration and places it under a `config` property in your initial redux
-state.
+configuration in `app.config.js` and places it under a `config` reducer in your initial redux
+state. Example below appears in `config` reducer in client-side code
```
kinetifex:
```suggestion
configuration in `app.config.js` and places it under a `config` property in your initial redux
state. Example below can selected from `state.config.url` in client-side code
```

PR #154, opened 2020-03-05
Allow context to set packageManager without overriding installCmd or localCmd

srapp:
## Summary

When setting a default `packageManager` in the `createContext` of a new plugin, the `installCmd` and `localCmd` values never get set, which lead to the Readme.md looking something like this:

Now start up the app.

```bash
cd process-manager-ui




```

## Changelog

Allow `createContext` to set `packageManager` without overriding `installCmd` or `localCmd`

## Test Plan

Added unit tests that fail without the code changes.  Totally open to finding a better way of testing locally with an actual `gasket create`


indexzero:
Thank you for your contribution @srapp!

PR #156, opened 2020-03-29
Unique names for anonymous plugins

apoco:
## Summary

I'm building a framework with `@gasket/engine`. When setting up the engine, I used the `plugins.add` array in the config, passing in plugin _objects_. Strangely, only the _first_ plugin did anything when invoking a hooked lifecycle. It turned out that this is because of how `@gasket/resolve` processes plugin objects. It requires unique names. If a `name` property is missing, it creates the name `gasket-plugin-undefined`. Therefore, if two plugins are missing a name property, they're given this same name. Then `@gasket/resolve` discards the "duplicate" silently.

It seems like a less surprising behavior that would have saved me tons of head scratching would have been to either:

1. Throw an error if a plugin is unnamed (I actually thought I was naming them, but I erroneously used an `id` property)
2. Allow unnamed plugins by assigning them distinct names.

I went with the nicer option 2, since IMO naming plugins if you don't care to refer to them in other contexts seems pointless, but option 1 is reasonable if we don't want to go my route.

## Changelog

- Generate unique names for anonymous plugins and presets

## Test Plan

Created unit tests that exercised the problem and demonstrate the fix.

### DIFF
--- a/packages/gasket-resolve/CHANGELOG.md
+++ b/packages/gasket-resolve/CHANGELOG.md
@@ -1,6 +1,6 @@
 # `@gasket/resolve`
 
-- Generate unique names for anonymous plugins and presets
+- Validate that plugin objects are named
 
 ### 5.1.0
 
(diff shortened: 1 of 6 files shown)


apoco:
@kinetifex: switched to using `throw` instead.

It looks like the plugin engine no-longer supports preset objects being passed in, only preset names, so I only needed a specific error message in the case of plugins.

PR #158, opened 2020-04-09
Drop in pwa

kinetifex:
## Summary

This PR automatically integrates the SW registration script for apps using the webpack plugin.

## Changelog
**@gasket/plugin-service-worker**
- Inject Wepack entries modules with SW registration script with the ability to configure which ones.

**@gasket/plugin-nextjs**
- Configure SW registration script injection to only the _app entry module.

## Test Plan

Local in app testing and added unit tests.


packages/gasket-plugin-service-worker/lib/utils.js
```diff
@@ -17,6 +17,32 @@ async function getCacheKeys(gasket) {
...
+ * Loads template file once with substitutions from config
+ *
+ * @param {Object} config - ServiceWorker config from gasket.config
+ * @returns {Promise<string>} template
+ */
+async function loadRegisterScript(config) {
+  if (!__script) {
+    const util = require('util');
+    const fs = require('fs');
+    const readFile = util.promisify(fs.readFile);
```
SivanMehta, line 32: Why can't you do this at the top of the file?

kinetifex: I opted for lazy-loading them here since they're only used once in the module and it makes the [test setup simple](https://github.com/godaddy/gasket/pull/158/files#diff-35e833b71eb2965283c3cc48cccff9beR62-R65). Concerns or just curious?

PR #163, opened 2020-06-02
Add an ASCII art logo to help output

tarkatronic:
## Summary

This adds an ASCII art version of the Gasket logo to the output of the `gasket` and `gasket help` CLI commands.

The artwork was generated here: https://asciiart.club/

## Changelog

Done!

```md
- An ASCII version of the Gasket logo is now displayed in the CLI help
```

## Test Plan

I was unsure of an good way to unit test this, since it's effectively not so much new functionality in this project, but relying on the new functionality in `oclif`.

But, I do have a screenshot!
gasket-ascii


SivanMehta:
Screen Shot 2020-06-02 at 1 16 08 PM

What are these extra 14,000 lines of code coming from?

tarkatronic:
> Screen Shot 2020-06-02 at 1 16 08 PM
> 
> What are these extra 14,000 lines of code coming from?

I honestly couldn't tell you 🤣 
All I did which should have touched that at all was
```
lerna add @oclif/plugin-help --scope=gasket/cli
npm ci
```
I didn't make any modifications directly to the top level `package.json`.

packages/gasket-cli/src/help.js
```diff
@@ -0,0 +1,22 @@
+const Help = require('@oclif/plugin-help').default;
+
+module.exports = class GasketHelp extends Help {
+  showHelp(args) {
+    console.log(`
```
SivanMehta, line 5: @3rd-Eden This logo already exists, but [only in the `create` lifecycle](https://github.com/godaddy/gasket/blob/master/packages/gasket-cli/src/scaffold/actions/print-report.js#L70-L75). Was the purpose of this work to have the logo run for *every* command? If so, shouldn't this PR also remove that bit of code?

3rd-Eden: Probably don't want to run it on **every** command, e.g. it has no point for `gasket start` which run in production environments, but it wouldn't hurt to display it during other development commands.

tarkatronic: Given that a text logo already exists elsewhere, we should probably use a single source of logo truth. So then two questions:

1. Which version of the logo should we use? (I prefer the new one, but I may be biased 😉 )
2. What other command(s) should display this? I only see `create` under the `commands/` folder, so I must be missing something if there is also a `start` command...

andyfleming: I think it would be overwhelming to have the logo output on every command. `create` feels like the appropriate time for it.

tarkatronic:
@indexzero @3rd-Eden @andyfleming @SivanMehta 
Howdy folks! I have made a few updates, after distilling your comments down.

The ascii logo now only shows up when the following commands are run:

1. `gasket`
2. `gasket help`
3. `gasket create`

It does not happen on any other commands, or on the help for individual commands.

Before I push these changes however, since the size of the logo was brought up as a potential issue, I wanted to show what this could look like with it reduced to 80% of the original size, in case that is less of a visual assault.

First, the current full size:
Annotation 2020-06-05 103309

Then, cut down to 80%:
Annotation 2020-06-05 103708

3rd-Eden:
Smaller version looks better IMHO.

PR #165, opened 2020-06-25
expose loadLocaleFilesIntoStore

chrisajimenez:
## Summary
Exposing loadLocaleFilesIntoStore so that intl strings can be populated outside of components.

### DIFF
--- a/packages/gasket-intl/README.md
+++ b/packages/gasket-intl/README.md
@@ -269,6 +269,34 @@ const performSomeAction = (payload) => {
 }
 ```
 
+### loadLocaleFilesIntoStore
+
+
+This function loops through the module names provided
+and reads the locale files for each of those identifiers and saves them in the Redux store.
+Typically, the `withLocaleRequired` hoc would handle this under the hood if you are working within components. If you need to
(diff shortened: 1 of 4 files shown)


packages/gasket-intl/README.md
```diff
@@ -269,6 +269,34 @@ const performSomeAction = (payload) => {
 }
 ```
 
+### loadLocaleFilesIntoStore
+
+
+This function loops through the module names provided
+and reads the locale files for each of those identifiers and saves them in the Redux store.
+Typically, the `withLocaleRequired` hoc would handle this under the hood if you are working within components. If you need to
```
SivanMehta, line 277:
```suggestion
Typically, `withLocaleRequired` would handle this under the hood if you are working within components, but if you need to
```

packages/gasket-intl/README.md
```diff
@@ -269,6 +269,34 @@ const performSomeAction = (payload) => {
 }
 ```
 
+### loadLocaleFilesIntoStore
+
+
```
tovwiv, line 274:
```suggestion
```

PR #167, opened 2020-07-17
first draft of starter pack

mramad2:
Goal is to reduce the barrier to entry for users of the framework by not requiring @gasket/cli to be globally installed. Using yarn create, npm init, or npx, a user can skip the global install and simply use one of these commands to spin up a gasket app.

### DIFF
--- a/packages/create-gasket-app/README.md
+++ b/packages/create-gasket-app/README.md
@@ -1,0 +1,31 @@
+# create-gasket-app
+
+Starter Pack for creating Gasket apps. 
+
+### Get Started Immediately
+
+You **don’t** need to install gasket-cli tools
+
+To create a new app, you may choose one of the following methods:
+
+### npx
+
+```sh
+npx create-gasket-app my-app
+```
+
+### npm
+
+```sh
+npm init gasket-app my-app
+```
+
+### Yarn
+
+```sh
+yarn create gasket-app my-app
+```
+
+## License
+
+[MIT](./LICENSE.md)
--- a/packages/create-gasket-app/README.md
+++ b/packages/create-gasket-app/README.md
@@ -4,7 +4,7 @@ Starter Pack for creating Gasket apps.
 
 ### Get Started Immediately
 
-You **don’t** need to install gasket-cli tools
+You **don’t** need to globally install the Gasket CLI
 
 To create a new app, you may choose one of the following methods:
 
--- a/packages/create-gasket-app/index.js
+++ b/packages/create-gasket-app/index.js
@@ -1,0 +1,10 @@
+#!/usr/bin/env node
+
+const { fork } = require('child_process');
+
+const [,, ...args] = process.argv;
+
+function main() {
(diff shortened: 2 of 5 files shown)


packages/create-gasket-app/README.md
```diff
@@ -0,0 +1,31 @@
+# create-gasket-app
+
+Starter Pack for creating Gasket apps. 
+
+### Get Started Immediately
+
+You **don’t** need to install gasket-cli tools
```
kinetifex, line 7:
```suggestion
You **don’t** need to globally install the Gasket CLI
```

packages/create-gasket-app/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"no test specified\" && exit 0",
+    "init": "./index.js"
+  },
+  "bin": {
+    "create-gasket-app": "./index.js"
+  },
+  "author": "",
+  "license": "ISC",
```
kinetifex, line 14: This should be MIT

packages/create-gasket-app/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "starter pack for creating a gasket app",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"no test specified\" && exit 0",
+    "init": "./index.js"
+  },
+  "bin": {
+    "create-gasket-app": "./index.js"
+  },
+  "author": "",
```
kinetifex, line 13:
```suggestion
  "author": "GoDaddy Operating Company, LLC",
```

packages/create-gasket-app/index.js
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env node
+
+const { fork } = require('child_process');
+
+const [,, ...args] = process.argv;
+
+function main() {
+  return fork(path.join(__dirname, 'node_modules', '.bin', 'gasket'), [...['create'], args], {stdio: 'inherit', stdin: 'inherit', stderr: 'inherit'});
```
SivanMehta, line 8: To me this doesn't look like it would work if `npm` hoists `@gasket/cli` to a directory other than something immediately colocated with this file. Given that there are no tests for this, how have you verified that this works?

kinetifex: @SivanMehta This is the path to the symlink bin file which npm will point to wherever it was installed/hoisted. You can see the path is not to an actual, i.e. `__dirname/node_modules/@gasket/cli/bin/run` which could change as you mention. So this seems safe.

mramad2: @SivanMehta It's working. That being said, I I should have indicated how I tested this when I submitted the PR.  

On a side note: I did encounter a bug when I ran a quick sanity test and updated PR with a fix.  :sweat_smile:

I tested via the following: 
Test 1). Using npm link. 
Fetch down project, and run `npm link` in project folder. You can now run the following command. **npm create-gasket-app my-app** on your machine. It will validate that create-gasket-app script calls underlying gasket_cli library correctly. One caveat is that the node_modules install fails when creating the gasket app. 

Test 2.) Published package to npm 
I renamed the package name, and published it. Tested npx and yarn scenarios.

PR #168, opened 2020-08-07
Do Not Autoload Test Files in Lifecycle Plugin

esundahl:
## Summary

Keeping test files in the lifecycle folder was causing them to be autoloaded. This PR filters out `*.spec.js` and `*.test.js` files allowing them to be kept alongside their corresponding modules.

## Changelog

- Updated Lifecycle plugin filter function
- Added addition tests

## Test Plan

Added basic unit tests for `some-event.test.js` and `some-event.spec.js`

Results of testing some basic file names.
```
express.js -> true
app.express.js ->  true
express.test.js ->  false
app.express.test.js ->  false
express.spec.js ->  false
app.express.spec.js ->  false
test.md ->  false
test.js ->  true
spec.js ->  true
test.asd.js ->  true
test.spec.js ->  false
index.html ->  false
special.js ->  true
special.spec.js ->  false
respec.js ->  true
asdftest.js ->  true
```

### DIFF
--- a/packages/gasket-plugin-lifecycle/index.js
+++ b/packages/gasket-plugin-lifecycle/index.js
@@ -26,7 +26,7 @@ async function resolve(root, name) {
     }
   }
 
-  const reg = /.*^((?!(test|spec).js).)*\.js$/;
+  const reg = /.*^((?!(\.test|\.spec).js).)*\.js$/;
   return files.filter(function filter(file) {
     return !!reg.exec(file);
   }).map(function each(file) {
--- a/packages/gasket-plugin-lifecycle/index.js
+++ b/packages/gasket-plugin-lifecycle/index.js
@@ -26,27 +26,27 @@ async function resolve(root, name) {
     }
   }
 
-  const reg = /.*^((?!(\.test|\.spec).js).)*\.js$/;
-  return files.filter(function filter(file) {
-    return !!reg.exec(file);
-  }).map(function each(file) {
-    const extname = path.extname(file);
-    const event = camelCase(path.basename(file, extname));
+  const isJs = /\.js$/i;
+  const isTest = /\.(spec|test)\./i;
+  return files.filter(file => isJs.test(file) && !isTest.test(file))
+    .map(function each(file) {
+      const extname = path.extname(file);
+      const event = camelCase(path.basename(file, extname));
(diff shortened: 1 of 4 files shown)


packages/gasket-plugin-lifecycle/index.js
```diff
@@ -26,8 +26,9 @@ async function resolve(root, name) {
     }
   }
 
+  const reg = /.*^((?!(\.test|\.spec).js).)*\.js$/;
```
SivanMehta, line 29: This assumes that the test files are name with the convention of `XXXXX.test.js` or `XXXX.spec.js`. It might be better to make this the default regex for a configurable value.

DullReferenceException, line 29: It's odd that you have a `^` anchor _after_ a `.*`, meaning that `.*` is never necessary. This should be equivalent:

```js
const isJs = /\.js$/i;
const isTest = /\.(spec|test)\./i;

return files
  .filter(file => isJs.test(file) && !isTest.test(file))
  .map(...)
```

packages/gasket-plugin-lifecycle/index.js
```diff
@@ -26,8 +26,9 @@ async function resolve(root, name) {
...
   return files.filter(function filter(file) {
-    return path.extname(file) === '.js';
+    return !!reg.exec(file);
```
DullReferenceException, line 31: You can use `.test` on a RegExp instead and avoid the `!!` (it'll work without the !! also)

packages/gasket-plugin-lifecycle/index.js
DullReferenceException, line 29: Should make regex(s) case-insensitive for file names.

packages/gasket-plugin-lifecycle/index.js
esundahl: Thanks!

packages/gasket-plugin-lifecycle/index.js
esundahl: Much simpler. Fixed

esundahl: This appears to be fixed using your changes above.

PR #173, opened 2020-09-10
Redux refresh

kinetifex:
## Summary

Major changes tuning Redux support up for the latest Next.js + next-redux-wrapper changes.

## Changelog

*@gasket/plugin-redux*
- Generate a redux/store.js with support for adding reducers via plugins
- Set process.env.GASKET_MAKE_STORE_FILE for shared package imports

*@gasket/plugin-nextjs*
- Generate pages/_app.s with next-redux-wrapper setup from store
- Generate a redux/store.js with next-redux-wrapper setup

*@gasket/redux*
- Support for adding an entry root reducer
- No longer provide a default make-store

## Test Plan

- Additional unit tests
- Local app testing

Local testing is a bit messy. This is what I did.

To test **create**:
```
gasket create my-app -p @gasket/nextjs --plugins @gasket/redux@file:/path/to/gasket/packages/gasket-plugin-redux,@gasket/nextjs@file:/path/to/gasket/packages/gasket-plugin-nextjs
```

To test **build**/**start** and avoid duplicate react errors...
in package.json, change `file:/` version to `^5.6.0` for the above plugins and reinstall
```
rm -rf node_modules
npm i
```
Then you can:
```
cp /path/to/gasket/packages/gasket-plugin-redux/lib/* node_modules/@gasket/plugin-redux/lib
cp /path/to/gasket/packages/gasket-redux/lib/* node_modules/@gasket/redux/lib
npm run build && npm run start -- --env local
```

Make sure to `npm run build` gasket-redux before copying!


packages/gasket-plugin-nextjs/generator/redux/redux/store.js
```diff
@@ -0,0 +1,17 @@
+const { configureMakeStore } = require('@gasket/redux');
+const { HYDRATE, createWrapper } = require('next-redux-wrapper');
+{{{reduxReducers.imports}}}
+
+// Basic hydrate reducer for next-redux-wrapper
+// @see: https://github.com/kirill-konshin/next-redux-wrapper#usage
+const rootReducer = (state, { type, payload }) => type === HYDRATE ? { ...state, ...payload } : state;
```
kinetifex, line 7: As a bit of information. I had started down the path of hiding this next-redux-wrapper boilerplate along these lines: https://github.com/godaddy/gasket/compare/redux-refresh...redux-refresh-alt

But it felt dirty. Yes, it makes the store a little nicer to read, but this is the only pro. By generating, the pros are it removes assumptions on future next-redux-wrapper changes, allows apps to upgrade/downgrade next-redux-wrapper at their leisure, and provides a starting point for tuning the HYDRATE reducer as needed.

I know we're trying to avoid boilerplate with Gasket, but this feels like the best approach for this situation to me. Open to discussion, however, if there are strong opinions otherwise.

DullReferenceException, line 7: This seems like a potential source of lots of head scratching. Someone dispatching to their store on server events might expect that state changes are going to be deeply, not shallowly merged and see existing state wiped out. Deep merges may also do unexpected things depending on the merge algorithm, so I get going the direction of having users define their own `rootReducer`.

It would be nice to raise some warnings somewhere in the documentation about this default `HYDRATE` handler since it's a bit counter-intuitive.

packages/gasket-redux/src/configure-make-store.js
```diff
@@ -4,37 +4,68 @@ import { createLogger } from 'redux-logger';
...
  * @param {Function[]} [options.enhancers] - Any additional enhancers
  * @param {Boolean} [options.logging] - logging is enabled by default. Passing false will disable logging completely.
  * @param {Function} [options.thunkMiddleware] - Optionally provide an extra argument for thunks
  * @param {Function} [postCreate] - Optional callback
- * @returns {function(*=): Store<{}>} makeStore
+ * @returns {MakeStore} makeStore
  */
 export default function configureMakeStore(
-  { reducers = {}, initialState = {}, middleware = [], enhancers = [f => f], logging = false, thunkMiddleware = thunk } = {},
+  {
```
DullReferenceException, line 46: If `rootReducer` was given a default implementation, wouldn't that make the upgrade much easier? As is, you'd have to change your store code so this is a breaking change. Seems like a default shallow merge reducer (or deep if we switch to that) would be helpful here, since someone upgrading from an earlier version probably isn't even using store dispatches on the server side props handler.

kinetifex: A default would only be needed/helpful if next-redux-wrapper@6 is installed and being used. I had started down that path but had concerns I mentioned in [this comment](https://github.com/godaddy/gasket/pull/173#discussion_r486776184). Having no `rootReducer` specified results in our current behavior. It's just a new option that can be used, and allows fulfillment of a requirement for next-redux-wrapper@6.

I'd like to keep `@gasket/redux` and `@gasket/plugin-redux` be `next` and `next-redux-wrapper` agnostic.

packages/gasket-plugin-nextjs/generator/redux/redux/store.js
kinetifex: 👍🏻 Yea, I know I head scratched on this change for awhile. I'm thinking this doc is really better suited for the nextjs plugin, since the redux plugin doesn't and shouldn't be concerned with next or next-redux-wrapper. I'll move it over and add an upgrade section or another doc to describe this better.

DullReferenceException (approved):

packages/gasket-plugin-nextjs/generator/redux/redux/store.js
axburgess-godaddy: I think that works

axburgess-godaddy (approved):

PR #176, opened 2020-09-14
Replace cross-fetch with local polyfill and ponyfill for node

mramad2:
## Summary

Next is now poly-filling fetch with SSR support using node-fetch. Following through with similar approach by switching up @gasket/fetch to use native (or poly-filled) fetch in the browser and a pony-fill for SSR. 

## Changelog

Change-log updates in files changed

## Test Plan

As we now switched from passing through cross-fetch to poly-filling window.fetch for browsers and pony-filling node-fetch for node, we've added tests to validate the following:

- Validating exports: Fetch, Request, Response, Headers
- Sending out Get and Post Requests
- Validating abort/cancel works accordingly

### DIFF
--- a/packages/gasket-fetch/browser.js
+++ b/packages/gasket-fetch/browser.js
@@ -1,0 +1,9 @@
+
+exports = window.fetch; // To import fetch from @gasket/fetch
+exports.default = window.fetch; // For TypeScript consumers without esModuleInterop.
+exports.fetch = window.fetch; // To import {fetch} from @gasket/fetch
+exports.Headers = window.Headers;
+exports.Request = window.Request;
+exports.Response = window.Response;
+
+module.exports = exports;
--- a/packages/gasket-fetch/browser.js
+++ b/packages/gasket-fetch/browser.js
@@ -1,7 +1,6 @@
 
 exports = window.fetch; // To import fetch from @gasket/fetch
 exports.default = window.fetch; // For TypeScript consumers without esModuleInterop.
-exports.fetch = window.fetch; // To import {fetch} from @gasket/fetch
 exports.Headers = window.Headers;
 exports.Request = window.Request;
 exports.Response = window.Response;
--- a/packages/gasket-fetch/browser.js
+++ b/packages/gasket-fetch/browser.js
@@ -5,5 +5,6 @@ exports.fetch = window.fetch; // To import {fetch} from @gasket/fetch
 exports.Headers = window.Headers;
 exports.Request = window.Request;
 exports.Response = window.Response;
+exports.AbortController = window.AbortController;
 
 module.exports = exports;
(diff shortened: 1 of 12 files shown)


packages/gasket-fetch/node.js
```diff
@@ -0,0 +1,2 @@
+const fetch = require('node-fetch');
```
smehta-godaddy, line 1: Won't this fail if consumers don't install `devDependencies`?

packages/gasket-fetch/browser.js
```diff
@@ -0,0 +1,9 @@
+
+exports = window.fetch; // To import fetch from @gasket/fetch
+exports.default = window.fetch; // For TypeScript consumers without esModuleInterop.
+exports.fetch = window.fetch; // To import {fetch} from @gasket/fetch
```
kinetifex, line 4: I don't think we need this. The API should for the browser module should be the same as for node, which does not support `{ fetch }` import and I don't think there's a case for it.

packages/gasket-fetch/browser.js
```diff
@@ -0,0 +1,9 @@
...
+exports.Headers = window.Headers;
+exports.Request = window.Request;
+exports.Response = window.Response;
+
```
kinetifex, line 8: Let's make `AbortController` available from this package in browser/node.

```suggestion
exports.AbortController = window.AbortController;

```

packages/gasket-fetch/package.json
```diff
@@ -23,7 +25,8 @@
     "url": "https://github.com/godaddy/gasket/issues"
   },
   "homepage": "https://github.com/godaddy/gasket/tree/master/packages/gasket-fetch#readme",
-  "dependencies": {
-    "cross-fetch": "^3.0.4"
+  "dependencies": {},
+  "devDependencies": {
+    "node-fetch": "^2.6.1"
```
kinetifex, line 30: Should be in `dependencies`

kinetifex (changes requested):

packages/gasket-fetch/node.js
mramad2: Yes, I didn't pay attention on the install. 🤦‍♂️

packages/gasket-fetch/test/browser.spec.js
```diff
@@ -0,0 +1,12 @@
+import 'whatwg-fetch';
+import { describe, it } from 'mocha';
+
+import fetch, { AbortController, Request, Headers, Response } from '../browser';
```
kinetifex, line 4: To be consistent, destructure these from the imported `fetch` like you did for the server.spec.js
```suggestion
import fetch from '../browser';
const { AbortController, Request, Headers, Response } = fetch;
```

kinetifex (approved):
Small tweak then 👍🏻

kawikabader (approved):

PR #179, opened 2020-10-01
Allow testing plugins to remove themselves after create lifecycle

rmarkins-godaddy:
## Summary

This PR solves the problem of adding `@gasket/plugin-mocha` or `@gasket/plugin-jest` and having it persist as a dependency, though it is no longer needed. Those plugins just add other packages to `devDependencies` and have no real use after that.

## Changelog

- Changelog updates in files changed

## Test Plan

Add tests around new remove method

### DIFF
--- a/packages/gasket-cli/test/unit/scaffold/config-builder.test.js
+++ b/packages/gasket-cli/test/unit/scaffold/config-builder.test.js
@@ -236,6 +236,30 @@ describe('ConfigBuilder', () => {
     });
   });
 
+  describe('.remove(key)', () => {
+    it('removes the desired package', () => {
+      pkg.add('dependencies', { 'some-pkg': '^1.2.0' });
+      assume(pkg.fields.dependencies['some-pkg']).to.exist();
+      pkg.remove('dependencies', 'some-pkg');
+      assume(pkg.fields.dependencies['some-pkg']).to.not.exist();
+    });
+
+    it('can remove multple dependencies', () => {
+      pkg.add('dependencies', { 'some-pkg': '^1.2.0', 'some-pkg2': '^1.1.0' });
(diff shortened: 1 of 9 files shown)


DullReferenceException:
Wouldn't this cause the plugin engine to complain because the preset still references the plugin package, or has that already been addressed?

rmarkins-godaddy:
> Wouldn't this cause the plugin engine to complain because the preset still references the plugin package, or has that already been addressed?

That's a good question. I could have sworn that was already addressed. @kinetifex do you know if this is still an issue?

kinetifex:
Yes, the plugin engine would fail to load due to the missing dependency. You need to remove the plugin from the package.json as well as the gasket.config.js. Which, for gasket.config, is a nested property your `remove` method will need to somehow support (`plugins.add: ['@gasket/jest']`).Also, these two files are already written out _before_ the `postCreate` lifecycle. So you'd have to write over them again if you're removing something _after_ they've been installed.

This is not a straight-forward problem to solve as it seems on the surface.

packages/gasket-cli/test/unit/scaffold/config-builder.test.js
```diff
@@ -236,6 +236,30 @@ describe('ConfigBuilder', () => {
...
 
+  describe('.remove(key)', () => {
+    it('removes the desired package', () => {
+      pkg.add('dependencies', { 'some-pkg': '^1.2.0' });
+      assume(pkg.fields.dependencies['some-pkg']).to.exist();
+      pkg.remove('dependencies', 'some-pkg');
+      assume(pkg.fields.dependencies['some-pkg']).to.not.exist();
+    });
+
+    it('can remove multple dependencies', () => {
```
jricaldi, line 247: TYPO. `multiple`

PR #180, opened 2020-10-09
Gasket Intl v2

kinetifex:
## Summary

Major updates to Gasket Intl support with support for latest Next.js data fetching functions and support for static site generation. This is effectively a re-write of the intl plugin and package.

## Changelog

**@gasket/intl**
- Decoupled from Redux
- Manifest is imported for bundling
- Support for `getStaticProps` and `getServerSideProps` in Next.js apps
  - Removed use of `getInitialProps` in components.
- Simplified locale paths (formerly referred to as "identifiers")

**@gasket/plugin-intl**
- Decoupled from Redux
  - Pertinent settings moved manifest
  - Request-based settings moved to `gasketData` in response object
- Decoupling from Express
  - Locale files should be served with apps static content
  - Middleware exist for handling locales on server-side
- Alignment of options terms to `locale` instead of `locale` + `language`

**@gasket/helper-intl**
- Common internal utils used by intl packages

**@gasket/plugin-nextjs**
- Generate `_app.js` with `@gasket/intl` and/or `@gasket/redux` if included

**@gasket/plugin-service-worker**
- Pass `res` to lifecycles

**@gasket/plugin-workbox**
- Pass `res` to lifecycles

## Test Plan

- Replace/add unit tests
- Local testing in-app


packages/gasket-plugin-intl/lib/index.js
```diff
@@ -1,102 +1,104 @@
...
 
-      app.use('/_locales', serveStatic(outputDir, {
-        index: false,
-        maxAge: '1y',
-        immutable: true
-      }));
+      return async function intlMiddleware(req, res, next) {
+        /* eslint-disable require-atomic-updates */
+        const acceptLanguage = (req.headers['accept-language'] || defaultLocale).split(',')[0];
+        const locale = await gasket.execWaterfall('intlLocale', acceptLanguage, req, res);
```
kinetifex, line 64: I changed `intlLanguage` to `intlLocale` to standardize our term for this. when moving this lifecycle out of `initReduxState`. I'm wondering if we should still retain `intlLanguage` but with a deprecated warning. Thoughts?

packages/gasket-intl/src/components.js
```diff
@@ -0,0 +1,259 @@
...
+
+/**
+ * Make an HOC that loads a locale file before rendering wrapped component
+ *
+ * @param {LocalePathPart} localesPath - Path containing locale files
+ * @param {object} [options] - Options
+ * @param {React.Component} [options.loading] - Custom component to show while loading
+ * @returns {function} wrapper
+ */
+export function withLocaleRequired(localesPath = manifest.localesPath, options = {}) {
```
msluther, line 236: Maybe consider breaking apart this file and having something like

`src/components/index.js` - export from all the others
`src/components/with-locale-required.js` - just this hoc
etc.

packages/gasket-intl/src/components.js
```diff
@@ -0,0 +1,259 @@
...
+  const { loading = null } = options;
+  /**
+   * Wrap the component
+   * @param {React.Component} Component - Component to wrap
+   * @returns {React.Component} wrapped component
+   */
+  return Component => {
```
msluther, line 243: consider using [`addhoc`](https://github.com/godaddy/addhoc) so that refs are forwarded, statics get hoisted, etc.

msluther (commented):
I've only made it through to `packages/gasket-intl/src/components.js` so far, will pick up again next week.

packages/gasket-intl/test/components.test.js
```diff
@@ -0,0 +1,152 @@
+import React from 'react';
+import assume from 'assume';
+import sinon from 'sinon';
+import proxyquire from 'proxyquire';
+import { mount } from 'enzyme';
+import { IntlProvider } from 'react-intl';
+import { ERROR, LOADED } from '../src/utils';
+
+const fetchStub = sinon.stub();
```
msluther, line 9: you should put this and the proxyquire call into the `beforeEach` for better test independence

packages/gasket-intl/test/components.test.js
```diff
@@ -0,0 +1,152 @@
...
+const mockManifest = require('./fixtures/mock-manifest.json');
+const mockConfig = {};
```
msluther, line 11: You're modifying this in tests (at least in the beforeEach, you should move the declaration into the `describe`, change it to a `let` and just make a new mockConfig for each test.

packages/gasket-intl/test/components.test.js
```diff
@@ -0,0 +1,152 @@
...
+      });
+    });
+  });
+
+  describe('LocaleRequired', function () {
+    let wrapper;
+
+    const doMount = props => mount(<mod.LocaleRequired { ...props }>MockComponent</mod.LocaleRequired>);
+
+    describe('#render', () => {
```
msluther, line 128: [Passing arrow functions (aka “lambdas”) to Mocha is discouraged.](https://mochajs.org/#arrow-functions)

packages/gasket-plugin-intl/lib/index.js
msluther: I'm okay with removing it since this will be a new major version anyway, but it should be documented clearly (is there a migration document for the new major version?)

packages/gasket-plugin-intl/lib/index.js
```diff
@@ -1,102 +1,105 @@
...
-      app.get('/locales-manifest.json', async (req, res) => {
-        const data = loadLocalesManifest(outputDir);
-        res.send(data);
-      });
+        // The gasketData object allows certain config data to be available for
+        // rendering as a global object for access in the browser.
+        res.gasketData = res.gasketData || {};
+
+        // TODO (@kinetifex): This could probably match LocalesProps used by Next.js loaders,
+        //   along with methods on req to preload locale files.
```
msluther, line 73: when? a new major version seems like a good time to make these sorts of changes.

packages/gasket-plugin-intl/package.json
```diff
@@ -4,11 +4,12 @@
   "description": "NodeJS script to build localization files.",
   "main": "lib",
   "scripts": {
-    "lint": "eslint lib/*.js *.js",
+    "lint": "eslint .",
     "lint:fix": "npm run lint -- --fix",
-    "test": "jest",
-    "test:watch": "jest --watchAll",
-    "test:coverage": "jest --coverage",
+    "test": "npm run test:runner",
```
msluther, line 9: any reason `test` doesn't run coverage by default?

packages/gasket-plugin-intl/test/build-manifest.test.js
```diff
@@ -0,0 +1,78 @@
...
+
+const writeFileStub = sinon.stub().callsFake((...args) => {
+  args[args.length - 1](null, true);
+});
+
+const buildManifest = proxyquire('../lib/build-manifest', {
+  fs: {
+    writeFile: writeFileStub
+  }
+});
```
msluther, line 14: IMHO, for better test independence, move these into the `beforeEach` so you get a brand new instance each time.

packages/gasket-plugin-intl/test/build-manifest.test.js
```diff
@@ -0,0 +1,78 @@
...
+          defaultLocale: 'en-US',
+          localesMap: {
+            'fr-CH': 'fr-FR'
+          },
+          localesDir: path.join(__dirname, 'fixtures', 'locales'),
+          manifestFilename: 'locales-manifest.json'
+        }
+      }
+    };
+  });
```
msluther, line 35: you should probably have a `sinon.restore()` in an `afterEach`

packages/gasket-plugin-intl/test/build-modules.test.js
```diff
@@ -0,0 +1,235 @@
...
+  describe('BuildModules', function () {
+    let builder;
+
+    beforeEach(function () {
+      builder = new BuildModules(mockGasket);
+    });
+
+    afterEach(function () {
+      builder = null;
+      sinon.restore();
```
msluther, line 41: IMHO, hoist that to an `afterEach` of the outermost `describe`

packages/gasket-plugin-intl/test/build-modules.test.js
```diff
@@ -0,0 +1,235 @@
...
+    describe('#copyFolder', function () {
+      it('calls saveJsonFile for correct data', function (done) {
+        sinon.stub(fs, 'mkdirp').returns(Promise.resolve());
+        sinon.stub(fs, 'readdir').resolves(['test/folder/name.json']);
+        sinon.stub(fsUtils, 'saveJsonFile').callsFake(() => {
+          assume(fs.mkdirp.getCalls()).length(1);
+          assume(fs.readdir.getCalls()).length(1);
+          assume(fs.readFile.getCalls()).length(1);
+          done();
+        });
```
msluther, line 59: why this over having `fsUtils.saveJsonFile` stubbed with `resolves()`, await the `builder.copyFolder` call and move the `assume`s after the await?

packages/gasket-plugin-intl/test/build-modules.test.js
```diff
@@ -0,0 +1,235 @@
...
+
+    it('sets up dir constants', function () {
+      assume(builder).property('_outputDir');
+      assume(builder).property('_nodeModulesDir');
+      assume(builder).property('_lookupDir');
+    });
+
+    describe('#copyFolder', function () {
+      it('calls saveJsonFile for correct data', function (done) {
+        sinon.stub(fs, 'mkdirp').returns(Promise.resolve());
```
msluther, line 52:
```suggestion
        sinon.stub(fs, 'mkdirp').resolves();
```

packages/gasket-plugin-intl/test/build-modules.test.js
```diff
@@ -0,0 +1,235 @@
...
+          done();
+        });
+        sinon.stub(fs, 'readFile').resolves('{ "key-1": "value-1" }');
+        builder.copyFolder(testSrcFilePath, testTgtFilePath);
+      });
+    });
+
+    describe('#processFiles', function () {
+      let fakeSrcDir, fakeTgtDir;
+      beforeEach(() => {
```
msluther, line 67: Per my other comment, you should not use arrow functions with mocha functions (it/describe/beforeeach/aftereach/etc)

packages/gasket-plugin-intl/test/build-modules.test.js
```diff
@@ -0,0 +1,235 @@
...
+        sinon.stub(builder, 'copyFile');
+        sinon.stub(builder, 'copyFolder');
+      });
+      it('calls copyFile for a json file', function () {
+        const fakeFileName = 'aa-AA.json';
+        const fakeFiles = [fakeFileName];
+
+        builder.processFiles(fakeSrcDir, fakeTgtDir, fakeFiles);
+
+        assume(builder.copyFile.getCalls()).length(1);
```
msluther, line 80: if you're using assume-sinon, you can do something like
```suggestion
        assume(builder.copyFile).was.called(1);
```

packages/gasket-plugin-intl/test/build-modules.test.js
```diff
@@ -0,0 +1,235 @@
...
+
+        sinon.stub(fs, 'lstat').callsFake(() => {
+          return new Promise((resolve) => {
+            resolve({
+              isDirectory: () => {
+                return true;
+              }
+            });
+          });
+        });
```
msluther, line 95:
```suggestion
        sinon.stub(fs, 'lstat').resolves({
          isDirectory: () => {
            return true;
          }
        });
```

packages/gasket-plugin-intl/test/build-modules.test.js
```diff
@@ -0,0 +1,235 @@
...
+
+    describe('#processDirs', function () {
+      let buildDirs;
+      beforeEach(function () {
+        buildDirs = [
+          '/src/test1/locales',
+          '/src/test2/locales'
+        ];
+        sinon.stub(fs, 'remove').returns(Promise.resolve());
+        sinon.stub(fs, 'mkdirp').returns(Promise.resolve());
```
msluther, line 140: why not use `resolves`?
```suggestion
        sinon.stub(fs, 'remove').resolves();
        sinon.stub(fs, 'mkdirp').resolves();
```

packages/gasket-plugin-intl/test/build-modules.test.js
```diff
@@ -0,0 +1,235 @@
...
+
+    describe('#discoverDirs', function () {
+      let discoveredDirs;
+      beforeEach(function () {
+        discoveredDirs = [
+          '/path/to/module/myh-fake',
+          '/path/to/module/myh-fake2',
+          '/path/to/module/myh-fake3'
+        ];
+        sinon.stub(fsUtils, 'getDirectories').callsFake(() => (discoveredDirs));
```
msluther, line 161:
```suggestion
        sinon.stub(fsUtils, 'getDirectories').returns(discoveredDirs);
```

msluther: or better yet since this should probably be async
```suggestion
        sinon.stub(fsUtils, 'getDirectories').resolves(discoveredDirs);
```

packages/gasket-plugin-intl/test/build-modules.test.js
```diff
@@ -0,0 +1,235 @@
...
+      });
+
+      it('returns a list of all locale paths', async function () {
+        sinon.stub(fs, 'lstat').callsFake(() => {
```
msluther, line 165: use `.resolves` here as in other places

packages/gasket-plugin-intl/test/build-modules.test.js
```diff
@@ -0,0 +1,235 @@
...
+          });
+        }).onCall(0).returns(
+          new Promise((resolve) => {
+            resolve({
+              isDirectory: () => {
+                return false;
+              }
+            });
+          })
+        );
```
msluther, line 195: use `.resolves`

packages/gasket-plugin-intl/test/build.test.js
```diff
@@ -0,0 +1,46 @@
...
+const proxyquire = require('proxyquire');
+
+const buildManifest = sinon.stub().resolves();
+const buildModules = sinon.stub().resolves();
+
+const plugin = proxyquire('../lib/index', {
+  './build-manifest': buildManifest,
+  './build-modules': buildModules
+});
+const { build: buildHook } = plugin.hooks;
```
msluther, line 12: move these into the outermost `beforeEach`

msluther (commented):
Here's the second half of my initial pass.

packages/gasket-intl/src/components.js
kinetifex: 👌🏻 Good catch. I failed to bring over using `hoist-non-react-statics` for these (still a package dep), but I'll check out addhoc first.

packages/gasket-plugin-intl/lib/index.js
kinetifex: Yes, I'm going to do this (and other todos) in a follow-up PR. Just trying to scope back this PR where I could.

packages/gasket-intl/README.md
```diff
@@ -10,291 +20,175 @@ npm i @gasket/intl
...
-import { LocaleRequired } from '@gasket/intl';
+const Component = props => <h1><FormattedMessage id='welcome'/></h1>
 
-<LocaleRequired>
-  <div>
-    <div><FormattedMessage id='app_more_info' /></div>
-    <div><FormattedMessage id='app_details' /></div>
-  </div>
-</LocaleRequired>
+export default withLocaleRequired('/locales')(Component);
```
salilagrawal, line 69: is the localesPath parameter mandatory?

packages/gasket-helper-intl/lib/index.js
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * Language code with region
+ * @typedef {Lang} Locale
+ *
+ * @example
+ * "en-US"
+ */
+
+const reLocale = /(\/[$:{]locale}?\/)/;
```
salilagrawal, line 45: Is this for checking the locale path param? name isn't clear enough I think.

packages/gasket-helper-intl/docs/api.md
```diff
@@ -0,0 +1,159 @@
...
+or a URL template with a `:locale` path param to a .json file.
+
+**Kind**: global typedef  
+**Example**  
+```js
+"/locales"
+```
+**Example**  
+```js
+"/locales/:locale/component.json"
```
salilagrawal, line 107: add a line with $locale also here if that's supported?

packages/gasket-intl/README.md
kinetifex: No. I add docs and an example describing using a default localesPath.

packages/gasket-plugin-intl/README.md
```diff
@@ -60,104 +60,242 @@ module.exports = {
...
+`en`, our fallback would look like:
+
+```
+de-CH -> de -> en-US -> en
+               └── (default)
+```
+
+So for `de-CH`, we would be loading the `en` locale file. Not ideal for your
+customers, but this serves as a safety mechanism to make sure your app remains
+somewhat readable for unexcepted locales. Also note, however, that you can
```
salilagrawal, line 175: unexpected (spelling)

packages/gasket-intl/README.md
```diff
@@ -69,6 +69,18 @@ const Component = props => <h1><FormattedMessage id='welcome'/></h1>
...
 ```
 
+If you only have a single locales path for your app, the default `localesPath`
+will be used if unspecified:
+
+```jsx
+export default withLocaleRequired()(Component);
+```
+
+This also the behavior for the other components and functions where
```
msluther, line 79:
```suggestion
This is also the behavior for the other components and functions where
```

packages/gasket-intl/src/locale-required.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @param {object} props - Props
+ * @param {LocalePathPart} props.localesPath - Path containing locale files
+ * @param {React.Component} [props.loading] - Custom component to show while loading
+ * @returns {JSX.Element|null} element
+ */
+export default function LocaleRequired(props) {
+  const { localesPath, loading = null, children } = props;
+  const loadState = useGasketIntl(localesPath);
+  // const loadState = 'useGasketIntl(localesPath)';
```
msluther, line 18: still need this comment?

packages/gasket-intl/src/with-intl-provider.js
```diff
@@ -0,0 +1,114 @@
+import React, { useReducer } from 'react';
+import hoistNonReactStatics from 'hoist-non-react-statics';
+import merge from 'lodash.merge';
+import { IntlProvider } from 'react-intl';
+import { GasketIntlContext } from './hooks';
```
msluther, line 5: Your context is part of the `hooks` file? Should it be somewhere else that both places can pull from?

packages/gasket-intl/src/with-intl-provider.js
```diff
@@ -0,0 +1,114 @@
...
+        ...state.messages,
+        [locale]: {
+          ...state.messages[locale],
+          ...messages
+        }
+      },
+      status: {
+        ...state.status,
+        [file]: LOADED
+      }
```
msluther, line 48: should the messages be separated by file? if there's 2 files with the same message key, are they going to stomp each other's message?

msluther: should this all be
```
return {
  ...state,
  [file]: {
     ...state[file],
     messages: {
        ...etc
     },
     status: LOADED/type/etc
   }
}
```

kinetifex: No, we need the combined messages for a locale, which is passed down to the react-intl provider. Yes, there is a risk of stomping over messages with the same key, but that seems like a rare edge case and more of a content problem for the app dev though. We could add a check and log a warning here, but it feels like static analysis of the JSON files would be most appropriate for teams. For example [eslint-plugin-i18n-json](https://github.com/godaddy/eslint-plugin-i18n-json) can do this.

packages/gasket-plugin-intl/README.md
```diff
@@ -60,104 +60,242 @@ module.exports = {
...
-## How it works
-
-When you build your app, the plugin looks for `locale` directories in the app
-root and in node modules. Each file is then read and output to the build or
-outputDir with a hash added to the filename. This hash is based on the contents
-of the locale file, so if there are any changes, the file will get a new hash to
-ensure it is unique.
+There are several strategies for managing locale files for an app. By default,
+the plugin assumes they will be static files committed to the app's under a
+`public/locales` directory.
```
salilagrawal, line 266: slightly confused...in other places it says that by default it will look into /locales folder for these files, and here it says public/locales. Also for other modules, it says the default is /locales. it should be /locales always, right? and the files should be copied over to public/locales on build?

packages/gasket-plugin-intl/README.md
```diff
@@ -60,104 +60,242 @@ module.exports = {
...
 
-Locale files are served with a long cache expiration time. This allows browsers
-to cache the locale files, and only versions with hash name changes will be
-newly downloaded.
+Another practice is to locale files under different npm modules. By enabling the
+`intl.modules` option in the `gasket.config.js`, when the app builds, the plugin
```
salilagrawal, line 269: Is the `intl.modules` option mandatory for the files to be picked up on build?

packages/gasket-plugin-intl/README.md
kinetifex: The files are disk would be in `./public/locales` in the repo by default, but served as `/locales` by the app. This is how [Next.js handles it](https://nextjs.org/docs/basic-features/static-file-serving).

If an app (non-Next.js) wants to serve the files under a different path, or store them on disk they can. Say the `localesDir` is `./static/locales` and served under `localesPath` of `/static/locales`.

I'll try to clarify this in the doc.

packages/gasket-plugin-intl/README.md
kinetifex: Yes, it's an opt-in feature now.

packages/gasket-plugin-intl/README.md
kinetifex: Ok let me know if the change in 6626ecd helps.

salilagrawal (approved):
Looks good!

packages/gasket-intl/src/with-intl-provider.js
msluther: ah okay. 👍

msluther (approved):

PR #181, opened 2020-10-14
Gasket API Preset

kawikabader:
## Summary

We would like to offer an API Preset for Gasket.

## Changelog

Changelog file included.
## Test Plan

The included plugins are sufficiently tested.


packages/gasket-plugin-express/index.js
```diff
@@ -60,7 +70,17 @@ module.exports = {
       await gasket.exec('express', app);
 
       const postRenderingStacks = await gasket.exec('errorMiddleware');
-      postRenderingStacks.forEach(stack => app.use(stack));
+      postRenderingStacks.forEach((stack) => app.use(stack));
+
+      if (config.express && config.express.routes) {
+        glob(`${config.express.routes}.js`, { cwd: gasket.config.root }, function (err, files) {
```
kinetifex, line 76: nit:
You use `gasket.config.root` in a couple of places. `config` is already destructured and it'd be a bit cleaner to destructure `root` as well up after line 44.

Same thing is true for `routes`. You could simplify your if check if you use this restructure trick:
```
const { root, express: { routes } = {} } = config;
```
If express is not set, default to an empty object. `routes` will be undefined of not otherwise set.

Now you can pass `root` and `routes` around, and change your if check to `if(routes)`.

kinetifex (approved):

burgessa23 (approved):

PR #183, opened 2020-10-20
Adding @gasket/data

kinetifex:
## Summary

New util package to read JSON data embedded in script tag in the browser.

## Changelog

**@gasket/data**
- Initial release

## Test Plan

- Unit tests


packages/gasket-data/README.md
```diff
@@ -0,0 +1,59 @@
...
+## Installation
+
+```
+npm i @gasket/data
+```
+
+## Usage
+
+This helper is intended for use in conjunction with Gasket Data embedded in a
+script tag in the html document. If the follow data is rendered...
```
DullReferenceException, line 14: following

kawikabader (approved):

packages/gasket-data/lib/index.js
```diff
@@ -0,0 +1,8 @@
+const deepFreeze = require('deep-freeze');
+
+let content = (document.getElementById('GasketData') || {}).textContent;
+if (content) {
+  content = deepFreeze(JSON.parse(content));
```
DullReferenceException, line 5: What benefits do we gain by freezing this data? I'm personally all for minimizing data mutation, but this could lead to unexpected runtime errors, not to mention it will add additional runtime overhead while traversing the data and running `Object.freeze`. Maybe this is too prescriptive.

kinetifex: Just trying to minimize abuse and set expectations that this should be read-only, used merely for transferring data from server to client.

It shouldn't be needed in production builds anyway. Maybe it's something we can conditionally set during development if we see it being abused.

I can leave it out for now.

DullReferenceException (approved):

PR #186, opened 2020-10-27
Added check for swagger.json file, with warning/error messaging if not found.

kawikabader:
## Summary

Currently, when a new Gasket API is created without any swagger config set in the gasket.config.js, devs are unable to start the service due to an error message being thrown.

## Changelog

Added check for swagger.json file, with warning/error messaging if not found. The Swagger `/api-docs` route will also display a generic error message: 
Screen Shot 2020-10-27 at 8 31 09 AM

## Test Plan

Tests have been updated.

### DIFF
--- a/packages/gasket-plugin-swagger/index.js
+++ b/packages/gasket-plugin-swagger/index.js
@@ -77,7 +77,7 @@ module.exports = {
 
         if (!swaggerSpec) {
           gasket.logger.warning(
-            `No JSDocs for Swagger were found in files (${jsdocs.apis}). Definition file not generated...`
+            `No JSDocs for Swagger were found in files (${jsdoc.apis}). Definition file not generated...`
           );
         } else {
           let content;
--- a/packages/gasket-plugin-swagger/index.js
+++ b/packages/gasket-plugin-swagger/index.js
@@ -77,7 +77,7 @@ module.exports = {
 
         if (!swaggerSpec) {
           gasket.logger.warning(
-            'Missing swagger config in gasket.config.js: swagger.json was not generated...'
+            `No JSDocs for Swagger were found in files (${jsdocs.apis}). Definition file not generated...`
           );
         } else {
           let content;
--- a/packages/gasket-plugin-swagger/index.js
+++ b/packages/gasket-plugin-swagger/index.js
@@ -5,6 +5,7 @@ const swaggerUi = require('swagger-ui-express');
 const swaggerJSDoc = require('swagger-jsdoc');
 const readFile = promisify(fs.readFile);
 const writeFile = promisify(fs.writeFile);
+const accessFile = promisify(fs.access);
 
 const isYaml = /\.ya?ml$/;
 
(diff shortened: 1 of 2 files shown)


packages/gasket-plugin-swagger/index.js
```diff
@@ -25,7 +25,13 @@ async function loadSwaggerSpec(root, definitionFile) {
       // eslint-disable-next-line require-atomic-updates
       __swaggerSpec = require('js-yaml').safeLoad(content);
     } else {
-      __swaggerSpec = require(target);
+      fs.access(target, fs.constants.F_OK, (err) => {
```
kinetifex, line 28: You should probably do this `fs.access` check before trying to read the file as `yaml`, not just if `json`.

packages/gasket-plugin-swagger/index.js
```diff
@@ -74,6 +80,10 @@ module.exports = {
 
         await writeFile(target, content, 'utf8');
         gasket.logger.info(`Wrote: ${definitionFile}`);
+      } else {
+        gasket.logger.warning(
```
kinetifex, line 84: This is a legitimate use case, and so should not warn here.

What we do need to warn on, is if `jsdoc` is configured but the swaggerSpec results are empty.

packages/gasket-plugin-swagger/index.js
```diff
@@ -89,6 +99,10 @@ module.exports = {
 
       const swaggerSpec = await loadSwaggerSpec(root, definitionFile);
 
+      if (!swaggerSpec) {
+        gasket.logger.error('Missing swagger.json file...');
```
kinetifex, line 103: User and change the `definitionFile` to be something other than `swagger.json` so use that in your warning.
```suggestion
        gasket.logger.error(`Missing ${definitionFile} file...`);
```

Also, it might be better to move this logging into the `loadSwaggerSpec`. You can pass the logger as an argument. That'll keep the logging paired with the loading code.

kinetifex (changes requested):

packages/gasket-plugin-swagger/index.js
kawikabader: @kinetifex am I correct in assuming that if the swaggerSpec results are empty, we also want to bypass https://github.com/godaddy/gasket/blob/UXP-4180/packages/gasket-plugin-swagger/index.js#L74-L82 ? Or should we simply warn and allow the rest of this block to run?

kinetifex: Yea, that seems to make sense to skip the writes if no content.

packages/gasket-plugin-swagger/index.js
```diff
@@ -65,15 +75,21 @@ module.exports = {
         const target = path.join(root, definitionFile);
         const swaggerSpec = swaggerJSDoc(jsdoc);
 
-        let content;
-        if (isYaml.test(definitionFile)) {
-          content = require('js-yaml').safeDump(swaggerSpec);
+        if (!swaggerSpec) {
+          gasket.logger.warning(
+            'Missing swagger config in gasket.config.js: swagger.json was not generated...'
```
kinetifex, line 80:
```suggestion
            `No JSDocs for Swagger were found in files (${jsdocs.apis}). Definition file not generated...`
```

kinetifex (approved):

axburgess-godaddy (approved):

burgessa23 (approved):

PR #214, opened 2020-11-10
[doc] Remove next-routes from README

blackswanny:
## Summary

since next-routes is not supported anymore, we should remove it to not confuse engineers


rxmarbles (approved):

kinetifex:
@blackswanny thanks for bringing this to our attention. Since this README is generated, the change needs to occur in the meta lifecycle for the plugin. I've made that change over in #215, so I am closing this one.

PR #218, opened 2020-11-30
Generate static web manifest

kawikabader:
## Summary

We would like the ability to create a web manifest on build of the application for use with static pages/apps.
## Changelog

Changelog to be updated upon approval.
## Test Plan

Tests have been updated.

### DIFF
--- a/packages/gasket-plugin-manifest/README.md
+++ b/packages/gasket-plugin-manifest/README.md
@@ -60,7 +60,7 @@ module.exports = {
 }
 ```
 
-If you want to generate a `manifest.json` file at build time for use with a static page/application, the plugin can be configured with the `staticOutput` option:
+If you want to generate a `manifest.json` file at build time for use with a static app, the plugin can be configured with the `staticOutput` option:
 
 ```js
 // gasket.config.js
--- a/packages/gasket-plugin-manifest/README.md
+++ b/packages/gasket-plugin-manifest/README.md
@@ -67,7 +67,7 @@ If you want to generate a `manifest.json` file at build time for use with a stat
 module.exports = {
   manifest: {
     // other options
-    staticOutput: '/custom/path/manifest.json' // recommended: /manifest.json
+    staticOutput: '/custom/path/manifest.json'
   }
 }
 ```
@@ -76,6 +76,8 @@ You will also need to include a link to your `manifest.json` file on your static
 ```html
 <link src="/manifest.json" rel="manifest">
 ```
+
+Users also have the option to pass in a boolean value of `true`, which will tell gasket to configure the output path to a default value of `/manifest.json`.
 
 ## Lifecycles
 
--- a/packages/gasket-plugin-manifest/README.md
+++ b/packages/gasket-plugin-manifest/README.md
@@ -60,6 +60,23 @@ module.exports = {
 }
 ```
 
+If you want to generate a `manifest.json` file at build time for use with a static page/application, the plugin can be configured with the `staticOutput` option:
+
+```js
+// gasket.config.js
+module.exports = {
+  manifest: {
+    // other options
+    staticOutput: '/custom/path/manifest.json' // recommended: /manifest.json
+  }
+}
+```
+You will also need to include a link to your `manifest.json` file on your static html pages:
+
+```html
+<link src="/manifest.json" rel="manifest">
+```
+
 ## Lifecycles
 
 ### manifest
--- a/packages/gasket-plugin-manifest/README.md
+++ b/packages/gasket-plugin-manifest/README.md
@@ -64,7 +64,7 @@ module.exports = {
 
 ### manifest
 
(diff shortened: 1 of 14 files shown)


packages/gasket-plugin-manifest/README.md
```diff
@@ -60,11 +60,30 @@ module.exports = {
...
+  }
+}
+```
+You will also need to include a link to your `manifest.json` file on your static html pages:
+
+```html
+<link src="/manifest.json" rel="manifest">
+```
+
+Users also have the option to pass in a boolean value of `true`, which will tell gasket to configure the output path to a default value of `/manifest.json`.
```
kinetifex, line 80: Correct the default path and make the description less verbose:
```suggestion
Users also have the option to pass in a boolean value of `true`, which defaults the path to `public/manifest.json`.
```

packages/gasket-plugin-manifest/README.md
```diff
@@ -60,11 +60,30 @@ module.exports = {
 }
 ```
 
+If you want to generate a `manifest.json` file at build time for use with a static page/application, the plugin can be configured with the `staticOutput` option:
```
kinetifex, line 63:
```suggestion
If you want to generate a `manifest.json` file at build time for use with a static app, the plugin can be configured with the `staticOutput` option:
```

packages/gasket-plugin-manifest/lib/configure.js
```diff
@@ -0,0 +1,20 @@
...
+ * @param {Gasket} gasket - Gasket
+ * @param {Object} config - Base gasket config
+ * @returns {Object} config
+ */
+module.exports = function configure(gasket, config = {}) {
+  const { config: { root } } = gasket;
+  const manifest = deepmerge(baseConfig, config.manifest);
+  const { staticOutput = false } = manifest;
+
+  manifest.staticOutput = prepareStaticOutputPath(staticOutput, root);
```
kinetifex, line 17: This `prepareStaticOutputPath` code is very simple and only used in this configure hook. Can you simply this by just inlining it here instead of extracting it to a separate util function?

packages/gasket-plugin-manifest/lib/express.js
```diff
@@ -0,0 +1,18 @@
...
+ * @param  {Gasket} gasket The gasket API
+ * @param  {Express} app gasket's express server
+ * @async
+ */
+async function express(gasket, app) {
+  const { config } = gasket;
+  const { staticOutput } = (config && config.manifest || {});
+  const { path } = (config && config.manifest || {});
+
+  app.get(staticOutput || path || baseConfig.path, (req, res) => {
```
kinetifex, line 13: `staticOutput` is where the file lives on disk (if outputted), and not the URL path.
```suggestion
  app.get(path || baseConfig.path, (req, res) => {
```

packages/gasket-plugin-manifest/lib/utils.js
```diff
@@ -0,0 +1,47 @@
...
+ * @returns {Object} manifest.json
+ */
+async function gatherManifestData(gasket, context) {
+  const { logger, execWaterfall, config } = gasket;
+  const source = (context.req && context.req.originalUrl) || 'static manifest';
+
+  logger.debug(`Gathering manifest for ${source}`);
+
+  const fromConfig = config.manifest;
+  const manifest = { ...baseConfig, ...fromConfig };
```
kinetifex, line 19: Make sure that the `staticOutput` property does not blend into the final `manifest.json` content. Add a test for this.

packages/gasket-plugin-manifest/lib/express.js
kinetifex, line 13: We only need a route if no `staticOutput`:

```suggestion
  if(!staticOutput) {
    app.get(path || baseConfig.path, (req, res) => {
```

packages/gasket-plugin-manifest/lib/middleware.js
```diff
@@ -0,0 +1,30 @@
...
+ * @param {Gasket} gasket - The gasket API
+ * @returns {function} Express middleware to apply
+ */
+function handler(gasket) {
+  const { serviceWorker: { url: swUrl = '' } = {} } = gasket.config;
+
+  const endpoints = [/manifest\.json/];
+  if (swUrl) endpoints.push(new RegExp(escapeRegex(swUrl)));
+
+  return async function manifestMiddleware(req, res, next) {
```
kinetifex, line 16: We should only add a middleware if no `staticOutput`.

You can exit early here:
```js
if(staticOutput) return;
```

kawikabader:
Updates from first round of PR feedback have been pushed. @kinetifex would you mind re-reviewing?

kinetifex (approved):

axburgess-godaddy (approved):

burgessa23 (approved):

Issue #224, opened 2020-12-10
Bring Back getInitialProps for Intl


kquerna:
## Description
Based on our latest testing, it seems like getInitialProps is more suitable for server-side rendering to load locale files than GSSP is. GSSP is called for client-routing too, which is unnecessary since locale files should only be fetched once. Let's bring back getInitialProps as an option, possibly again as part of the HOC.

**Package(s)**: @gasket/intl
**Branch**: canary

## AC:
* Ability to use getInitialProps using the withLocaleRequired

kinetifex:
Fixed in #225

PR #225, opened 2020-12-10
Opt-in Intl support for getInitialProps

kinetifex:
## Summary

Per #224, upon implementing `getServerSideProps` for intl, we found it to not be as performant as the previous `getInitialProps` implementation. The reason being is that with `getServerSideProps`, props are fetched with each browser route to the page. This is wasteful since locale files should really only be loaded once.

This introduces an opt-in `initialProps` option to the HOC, allowing `getInitialProps` to be added for Next.js pages which will preload locale files during SSR.

## Changelog

**@gasket/react-intl**
- Add `initialProps` options for enabled getInitialProps to preload locale files

**@gasket/plugin-intl**
- Attach the configured `localesDir` to `res.locals` for use during SSR

## Test Plan

- Local testing in dummy app
- Updated unit tests

### DIFF
--- a/packages/gasket-plugin-intl/lib/middleware.js
+++ b/packages/gasket-plugin-intl/lib/middleware.js
@@ -75,6 +75,8 @@ module.exports = function middlewareHook(gasket) {
       return messages[id] || defaultMessage || id;
     };
 
+    res.locals.localesDir = localesDir;
+
     next();
   };
 };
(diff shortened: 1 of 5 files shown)


DullReferenceException:
> ... with getServerSideProps, props are fetched with each browser route to the page. This is wasteful since locale files should really only be loaded once.

Can you explain what you mean by this? For both `getInitialProps` and `getServerSideProps`, this will only happen for full page loads, right? It'll still only be loaded once if you're doing client navigation, right?

kawikabader (approved):

burgessa23 (approved):

PR #231, opened 2020-12-17
App config available from @gasket/data

kawikabader:
## Summary

We need a way for app config to be made available in the browser, when Redux is not used. Today, teams can set config under a Redux property which will make it available in the Redux store.

For this change, we introduced a new special property (public) which will add its values to res.locals.gasketData.config, and thus be made available in the browser under with @gasket/data as gasketData.config.

## Test Plan

Tests updated.

### DIFF
--- a/packages/gasket-plugin-config/README.md
+++ b/packages/gasket-plugin-config/README.md
@@ -169,6 +169,19 @@ module.exports = {
 };
 ```
 
+### Config with Public config
+
+If you are not using Redux, but still need access to config values in client-side code, you can define a `public` property in your `gasket.config.js`. The config plugin will return these `public` properties to your browser, to be accessed by the `@gasket/data` plugin.
+
+```js
+module.exports = {
+  public: {
+    test1: 'config values here',
+    test2: 'config values here'
+  }
+};
+```
+
 ## Lifecycles
 
 ### appEnvConfig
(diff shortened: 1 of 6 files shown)


packages/gasket-plugin-config/README.md
```diff
@@ -169,6 +169,26 @@ module.exports = {
...
+
+```js
+module.exports = {
+  public: {
+    test1: 'config value 1 here',
+    test2: 'config value 2 here'
+  }
+};
+```
+The config plugin will return these `public` properties to your browser, to be accessed by the `@gasket/data` plugin, and used like so:
```
kinetifex, line 184: Refer to `@gasket/data` as a package or module, to distinguish it from a Gasket plugin (with lifecycle hooks, etc)
```suggestion
The config plugin will return these `public` properties to your browser, to be accessed by `@gasket/data`, and used like so:
```

kinetifex: Also, describe that what's placed in `public` in the config files, is accessible as `.config` from gasketData. You show that in the example, but should describe it.

kinetifex (changes requested):

kinetifex (approved):

mswaagman-godaddy (approved):

PR #254, opened 2021-02-17
Fix workbox setup for intl plugin

kinetifex:
## Summary

Uncovered a bug in the webpack hook for the intl plugin.

## Changelog

**@gasket/plugin-intl**
- Fix to use `defaultPath` in workbox lifecycle

## Test Plan

- Uncovered and tested in the canary app
- Unit tests


DullReferenceException (approved):
Seems to match documentation now. Not clear why it's named `defaultPath` though.

rmarkins-godaddy (approved):

PR #263, opened 2021-03-13
Protect against accesses of `res.locals.gasketData` when not defined

DullReferenceException:
When a static page is built, even the 404 page, it uses `_app.js`. If this app is using a `@gasket/react-intl` HOC with `initialProps: true`, it attempts to read the locale from `gasketData` which is undefined.


kinetifex (approved):

kawikabader (approved):

kinetifex:
This stemmed from a Next.js bug but is still a reasonable safety check to add.

PR #274, opened 2021-04-07
Improve logging for missing locale files

kinetifex:
## Summary

There were difficulties debugging when no locale files were found, or the locales-manifest.json failed to write (generally due to missing locale dir). This improves the logging for these cases.

## Changelog

**@gasket/plugin-intl**
- Improve logging for missing locale files and manifest write errors

## Test Plan

- Local app tested
- Added unit tests

### DIFF
--- a/packages/gasket-plugin-intl/lib/build-manifest.js
+++ b/packages/gasket-plugin-intl/lib/build-manifest.js
@@ -26,6 +26,10 @@ module.exports = async function buildManifest(gasket) {
   const files = (await glob('**/*.json', { cwd: localesDir }))
     .filter(f => f !== manifestFilename);
 
+  if (!files.length) {
+    logger.warning(`build:locales: No locale files found (${localesDir}).`);
+  }
+
   // generate a content hash for each file
   const paths = (
     await Promise.all(files.map(async file => {
@@ -54,6 +58,11 @@ module.exports = async function buildManifest(gasket) {
     paths
   };
 
-  await writeFile(tgtFile, JSON.stringify(manifest), 'utf-8');
-  logger.log(`build:locales: Wrote locales manifest.`);
+  try {
+    await writeFile(tgtFile, JSON.stringify(manifest), 'utf-8');
+    logger.log('build:locales: Wrote locales manifest.');
+  } catch (err) {
+    logger.error('build:locales: Unable to write locales manifest.');
+    throw err;
+  }
 };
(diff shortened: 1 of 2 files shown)


packages/gasket-plugin-intl/lib/build-manifest.js
```diff
@@ -54,6 +58,11 @@ module.exports = async function buildManifest(gasket) {
     paths
   };
 
-  await writeFile(tgtFile, JSON.stringify(manifest), 'utf-8');
-  logger.log(`build:locales: Wrote locales manifest.`);
+  try {
+    await writeFile(tgtFile, JSON.stringify(manifest), 'utf-8');
+    logger.log('build:locales: Wrote locales manifest.');
+  } catch (err) {
+    logger.error('build:locales: Unable to write locales manifest.');
```
DullReferenceException, line 65: Could you include `err.stack` here as well? That should include the error message and stack trace so it's clear why it failed. Or maybe `logger.error` has a cleaner way of logging errors.

kinetifex: So the stack will be displayed when the error is thrown still. There's not much in the stack however, which is I added the extra error logging. For example, with the logging changes:

```shell
» gasket build                                                          
 ›   Warning: No env specified, falling back to "development".
warning: build:locales: No locale files found (/path/to/app/public/locales).
error: build:locales: Unable to write locales manifest.
(node:49237) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open '/path/to/app/public/locales/locales-manifest.json'
(Use `node --trace-warnings ...` to show where the warning was created)
(node:49237) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
(node:49237) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
    Error: ENOENT: no such file or directory, open 
    '/path/to/app/public/locales/locales-manifest.json'
    Code: ENOENT

```

DullReferenceException (approved):

kawikabader (approved):

PR #275, opened 2021-04-09
Added --legacy-peer-deps flag to install cli.

kawikabader:
## Summary

The `create` command was failing when using npm7.

## Changelog

Added `--legacy-peer-deps` flag to install cli.
## Test Plan

Manually tested using npm6 and npm7.

### DIFF
--- a/packages/gasket-utils/lib/package-manager.js
+++ b/packages/gasket-utils/lib/package-manager.js
@@ -139,7 +139,9 @@ class PackageManager {
    * @public
    */
   async install(args = []) {
-    return this.exec('install', args);
+    // Installing with --legacy-peer-deps flag to accommodate npm7, specifically
+    // requiring different versions of react
+    return this.exec('install', [...args, '--legacy-peer-deps']);
   }
 
   /**
--- a/packages/gasket-utils/test/package-manager.test.js
+++ b/packages/gasket-utils/test/package-manager.test.js
@@ -70,6 +70,13 @@ describe('packageManager', function () {
         assume(res).is.a('object');
         assume(res.stdout).equals(stdout);
       });
+
+      it('contains legacy-peer-deps flag', async function () {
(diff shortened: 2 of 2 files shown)


burgessa23 (approved):

kinetifex (approved):

packages/gasket-utils/lib/package-manager.js
```diff
@@ -139,7 +139,9 @@ class PackageManager {
    * @public
    */
   async install(args = []) {
-    return this.exec('install', args);
+    // Installing with --legacy-peer-deps flag to accommodate npm7, specifically
+    // requiring different versions of react
+    return this.exec('install', [...args, '--legacy-peer-deps']);
```
rmarkins-godaddy, line 144: is this backwards comapatible? aka if i did this w/ `npm@6` will it still add `--legacy-peer-deps`?

rmarkins-godaddy: Should we also add tests around this?

rmarkins-godaddy:
@kinetifex @kawikabader did we try `--force` first prior to `--legacy-peer-deps`. I believe that would be more backward compat if someone was to use `npm@6` to use `gasket/cli`

packages/gasket-utils/lib/package-manager.js
kinetifex: `--force` has other implications, whereas `--legacy-peer-deps` handles the specific issue at hand.

kawikabader: > is this backwards comapatible? aka if i did this w/ `npm@6` will it still add `--legacy-peer-deps`?. iirc `npm@6` does not have this method. If you want to go this route you will want to check the `npm` version in use prior to this return

I did test this with npm@6 and had no issues installing.

rmarkins-godaddy (approved):

Issue #269, opened 2021-04-01
Load translations for defaultLocale for unsupported locales


kinetifex:
## Description
Currently, the [@gasket/plugin-intl] allows configuring a supported `locales` array, for the purpose of passing along to Next.js for i18n routing of static pages. A suitable requested feature is to also use this for other non-Next static pages. In the [@gasket/helper-intl], we should fallback back to the `defaultLocale` when supported `locales` are specified and the requested locale is not included.

**Package(s)**: @gasket/helper-intl

AC:
- When supported `locales` is configured, but requested locale is not included, load path for `defaultLocale`.

[@gasket/plugin-intl]:https://github.com/godaddy/gasket/tree/master/packages/gasket-plugin-intl
[@gasket/helper-intl]:https://github.com/godaddy/gasket/tree/master/packages/gasket-helper-intl

PR #276, opened 2021-04-20
Added check for locales config to determine fallbackLocale.

kawikabader:
## Summary

Currently, the @gasket/plugin-intl allows configuring a supported locales array, for the purpose of passing along to Next.js for i18n routing of static pages. A suitable requested feature is to also use this for other non-Next static pages. In the @gasket/helper-intl, we should fallback to the defaultLocale when supported locales are specified and the requested locale is not included.

https://github.com/godaddy/gasket/issues/269

## Changelog

Added check for locales config to determine fallbackLocale.
## Test Plan

Updated unit tests.

### DIFF
--- a/packages/gasket-helper-intl/lib/index.js
+++ b/packages/gasket-helper-intl/lib/index.js
@@ -95,7 +95,7 @@ const trim = localePath => localePath.replace(reLeadingSlash, '');
 function LocaleUtils(config) {
   const { manifest } = config;
   const { basePath = manifest.basePath } = config;
-  const { defaultLocale = 'en', localesMap, paths } = manifest;
+  const { defaultLocale = 'en', localesMap, paths, locales } = manifest;
   const defaultLang = defaultLocale.split('-')[0];
 
   /**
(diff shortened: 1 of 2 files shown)


packages/gasket-helper-intl/lib/index.js
```diff
@@ -149,6 +149,10 @@ function LocaleUtils(config) {
     const mappedLocale = localesMap && localesMap[locale] || locale;
     let fallbackLocale = mappedLocale;
 
+    if (locales && !locales[locale]) {
```
kinetifex, line 152: Locales should be an array, not a map
```suggestion
    if (locales && !locales.includes(locale)) {
```

Add a test the ensure that it does NOT fallback to default if requested locale IS in locales list.

kinetifex (changes requested):

kinetifex (approved):

burgessa23 (approved):

packages/gasket-helper-intl/lib/index.js
```diff
@@ -149,6 +149,10 @@ function LocaleUtils(config) {
     const mappedLocale = localesMap && localesMap[locale] || locale;
     let fallbackLocale = mappedLocale;
 
+    if (locales && !locales.includes(locale)) {
```
DullReferenceException, line 152: Shouldn't this be checking if `mappedLocale` is in `locales`?

kawikabader: Yep! Great catch!

DullReferenceException (approved):

PR #279, opened 2021-04-30
Make @gasket/plugin-log and @gasket/log slightly more flexible

msluther:
Without doing any breaking changes to rework how logging is done for gasket apps, this felt like the a decent approach to accomplish what I needed in my gasket application.

## Summary

I've made `@gasket/log` allow for sending in both custom `levels` and `format` properties for winston configuration. I've made `@gasket/log` also expose out some additional static properties so that the original/default values can be used to combine however the user wishes.

The issue I was trying to solve was that I needed to have an additional custom level on top of what gasket provided out of the box and also some additional custom formatters. I could have made this be `additionalLevels` and `additionalFormat` and internally done the combine work, but this seemed more flexible and a better API, but I'm open to something like that as well.

## Changelog

```md
- Allow for `levels` and `format` to be set on `@gasket/log`
```

## Test Plan

Existing tests all pass. I've added some additional tests to verify that `levels` and `format` were used appropriately if provided.

### DIFF
--- a/packages/gasket-log/server.js
+++ b/packages/gasket-log/server.js
@@ -66,26 +66,7 @@ class Log {
    * @private
    */
   format() {
-    if (this.local) {
-      return format.combine(
-        reduxLogger(),
-        format.colorize(),
-        format.splat(),
-        format.simple()
-      );
-    }
-
-    //
-    // Remark (@indexzero): in the event that redux logging is not
-    // turned off (i.e. configureMakeStore({ reducers })) AND level
-    // is manually set to 'debug' not having `inlineColors` here
-    // may cause issues.
-    //
-    return format.combine(
-      format.splat(),
-      format.label({ label: this.prefix }),
-      format.json()
-    );
+    return this.options.format || Log.getDefaultFormat(this.local, this.prefix);
   }
 
   /**
@@ -101,6 +82,10 @@ class Log {
     }
 
     return [new transports.Console()];
+  }
+
+  get levels() {
+    return this.options.levels || Log.levels;
   }
 
   /**
@@ -138,8 +123,9 @@ class Log {
    * @public
    */
   spawn() {
+    const levels = this.levels;
     const winston = createLogger({
-      levels: config.syslog.levels,
+      levels,
       level: this.level,
       silent: this.silent,
       format: this.format(),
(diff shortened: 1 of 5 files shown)


DullReferenceException (approved):

packages/gasket-log/server.js
```diff
@@ -149,16 +135,16 @@ class Log {
     //
     // Create proxies per loglevel to each method in winston.
     //
-    Object.keys(config.syslog.levels).forEach(level => {
-      this[level] = winston[level].bind(this.winston);
+    Object.keys(levels).forEach(level => {
+      this[level] = winston[level].bind(winston);
```
kinetifex, line 139: So, I just realized a potential problem with custom log levels... several of our plugins are pre-wired with logging at certain levels. What happens with `warning` [for example](https://github.com/godaddy/gasket/blob/6c37347cfe372506ea21af8ab340f7b1ac15c05f/packages/gasket-plugin-nextjs/config.js#L19)) if custom log levels are set to npm levels, which has `warn` instead?

DullReferenceException: Great point. Maybe better if these merged in with the default logging levels instead or with a set of expected logging levels.

kinetifex: Possibly? Would merging throw off prioritization though? Referencing [these levels](https://www.npmjs.com/package/winston#logging-levels), syslog `warning` is 4, but that's `verbose` for npm levels. Not sure that matters.

Another thought is maybe our plugins use a select that that must be provided, or can be auto resolved. For example `warning` could resolve to `warn`, but your custom log levels _must_ provide one or the other.

kinetifex (changes requested):
Blocking on this until we resolve [comments](https://github.com/godaddy/gasket/pull/279/files#r624160739).

packages/gasket-log/server.js
msluther: We could, however, just make sure every config.syslog.level is included in `levels` if there's a `levels` provided and just throw with a good message if there's something missing, then the caller is forced to decide what they want their various levels to mean.

msluther: I went with the "make sure the expected levels are defined" approach and put that in 47c6f2d

kinetifex (approved):

kawikabader (approved):

PR #280, opened 2021-05-04
Disable Next.js Telemetry

kawikabader:
## Summary

NextJS has a telemetry "feature" that anonymously sends tracking to Vercel about NextJS usage. (https://nextjs.org/telemetry). We would like to have Gasket apps opted-out of telemetry.

## Changelog

Updates to build lifecycle to disable Next.js telemetry.

## Test Plan

Unit tests added.


packages/gasket-plugin-nextjs/lib/build.js
```diff
@@ -0,0 +1,42 @@
...
+ */
+async function checkTelemetryStatus(logger) {
+  const statusRegex = /(?<=Status: ).*./;
+  const { stdout = '' } = await exec('npx next telemetry status');
+  const telemetryStatus = stdout.match(statusRegex)[0] || '';
+
+  if (telemetryStatus === 'Enabled') {
+    const { stdout: disableCmdOutput = '' } = await exec('npx next telemetry disable');
+    logger.info(`Next.js Telemetry: ${disableCmdOutput.match(statusRegex)[0]}`);
+  }
```
msluther, line 23: is disabling slow vs the status check? should we just always run disable?

Can we make use fo the env variable instead and just set `NEXT_TELEMETRY_DISABLED=1`

packages/gasket-plugin-nextjs/lib/build.js
```diff
@@ -0,0 +1,42 @@
...
+const { createConfig } = require('./config');
+
+/**
+ * Checks if Next.js Telemetry is enabled. If so, it disables it.
+ *
+ * https://nextjs.org/telemetry#how-do-i-opt-out
+ *
+ * @param {object} logger gasket logger
+ */
+async function checkTelemetryStatus(logger) {
```
msluther, line 15: naming nit:  The name implies that it's just checking, but it does in-fact disable it. maybe `ensure` instead of `check`?

packages/gasket-plugin-nextjs/lib/build.js
```diff
@@ -0,0 +1,42 @@
...
+  const statusRegex = /(?<=Status: ).*./;
+  const { stdout = '' } = await exec('npx next telemetry status');
```
msluther, line 17: is `npx` okay for gasket? will `yarn` users be okay? I know @kinetifex did some work a while back to make sure yarn worked.

decompil3d:
Why are we turning this off? Should this be something we disable in internal presets but leave on for open-source usage?

packages/gasket-plugin-nextjs/lib/build.js
kawikabader: @msluther I haven't seen or read about any sort of performance impact if we were to just run disable, so that may be a better/simpler direction here.

I think the env variable direction may actually be a better option.

packages/gasket-plugin-nextjs/lib/build.js
kawikabader: Great feedback. Will update.

packages/gasket-plugin-nextjs/lib/build.js
kawikabader: Really solid point. Will investigate further.

kawikabader:
> Why are we turning this off? Should this be something we disable in internal presets but leave on for open-source usage?

This is a really great suggestion. Will research this further.

Issue #211, opened 2020-10-30
Alternative to webpack-merge smart


kquerna:
## Description
The smart method of webpack-merge has been deprecated. We should move away from it in the webpack plugin. Reference the [webpack-merge docs](https://github.com/survivejs/webpack-merge) and [changelog](https://github.com/survivejs/webpack-merge/blob/master/CHANGELOG.md#503--2020-07-06) for details and suggested alternatives.

**Package(s)**: @gasket/plugin-webpack

AC:
- No longer merging webpack configs with deprecated method
- Merge results should match our existing output

Stretch:
- Ability for apps to customize webpack merge routine

DullReferenceException:
As brought up in [this thread](https://github.secureserver.net/gasket/gasket/pull/226), maybe our best bet is to use the gaskety waterfall approach, like:

```js
const webpackConfig = gasket.execWaterfallSync('webpack', nextWebpackConfig, data);
```

We could potentially pass the `webpack-merge` API and recommend using that so we wouldn't end up with a bunch of plugins all using different, possibly broken merging techniques. We should also consider obsoleting the `webpack-chain` technique and just have the one hook method.

It's also possible for us to introduce a waterfall merging technique without it being a breaking change. Leave the existing hook stuff which could be deprecated at a future time and add a new lifecycle at the very end, like maybe `webpackMerge`:

```js
const webpackConfig = gasket.execWaterfallSync('webpackMerge', currentConfig, data, webpackMerge);
```

That way applications and newer plugins could switch to the `webpackMerge` lifecycle without changing the semantics of existing hooks.

PR #283, opened 2021-06-16
Fix webpack lifecycle example

kinetifex:


kawikabader (approved):

3rd-Eden:
Given that both options have their uses, e.g. modifying existing configuration, wouldn't it make sense to put both examples in there?

rmarkins-godaddy (approved):

kinetifex:
@3rd-Eden I don't think mutating the initial config something we want to recommend right now. We have a ticket https://github.com/godaddy/gasket/issues/211 to provide alternates to the merge. Possibly a `webpackFinal` lifecycle, or some such name, that uses [execWaterfallSync](https://github.com/godaddy/gasket/blob/master/packages/gasket-engine/README.md#execwaterfallsyncevent-value-args) which will allow ordered modification of the full config object.

PR #284, opened 2021-06-16
New webpackConfig lifecycle

DullReferenceException:
## Summary

Add a `webpackConfig` lifecycle that gives users more control on how they want to merge webpack mutations in with the existing config.

Currently a user must formulate a config fragment that relies on a "smart" merge from `webpack-merge`, but that smart merge feature is going away. With this new lifecycle, users are passed a `webpack-merge` instance so they can use whatever techniques for merging they prefer or even do the merge through plain JavaScript.

### DIFF
--- a/packages/gasket-plugin-webpack/README.md
+++ b/packages/gasket-plugin-webpack/README.md
@@ -131,8 +131,8 @@ Executed after `webpack-chain` and `webpack`, it receives four parameters:
 1. The gasket API
 2. A webpack config object
 3. An options object with properties:
-   * `webpack` - The webpack API
-   * `nextOptions` - The [next.js webpack config options](https://nextjs.org/docs/api-reference/next.config.js/custom-webpack-config)
+   * `webpack` - The webpack API.
+   * `context` - Additional context about the webpack configuration being generated. For example, in next.js apps, this is the [next.js webpack config options](https://nextjs.org/docs/api-reference/next.config.js/custom-webpack-config).
    * `webpackMerge` - `The [`webpack-merge`](https://github.com/survivejs/webpack-merge/tree/v4.2.2) API, version 4.
 
 A hook should return a new webpack config object derived from the original. The usage of the `webpack-merge` API is recommended when doing so since properly handling the overloaded types within webpack config properties can be tricky. We recommend avoiding `webpack-merge` methods that have been deprecated in version 5 since a future version of this plugin may update to a new breaking version of `webpack-merge`.
@@ -142,7 +142,7 @@ A hook should return a new webpack config object derived from the original. The 
 function webpackHook(
   gasket,
   config,
-  { nextOptions: { isServer }, webpack, webpackMerge }
+  { context: { isServer }, webpack, webpackMerge }
 ) {
   return isServer
     ? config
--- a/packages/gasket-plugin-webpack/README.md
+++ b/packages/gasket-plugin-webpack/README.md
@@ -130,10 +130,11 @@ Executed after `webpack-chain` and `webpack`, it receives four parameters:
 
 1. The gasket API
 2. A webpack config object
-3. An options object with properties:
+3. A context object with the following properties:
    * `webpack` - The webpack API.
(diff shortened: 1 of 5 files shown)


packages/gasket-plugin-webpack/index.js
```diff
@@ -6,29 +6,37 @@ const { name, devDependencies } = require('./package');
...
+    ...configPartials
+  );
 
-  return webpackMerge.smart(webpackConfig, ...configs);
+  return execWaterfallSync(
+    'webpackMerge',
+    mergedConfig,
+    data,
+    webpackMerge
+  );
```
kinetifex, line 39: I like what this is doing, but I'm hung up on the naming. This is not really "merging", like the original lifecycle does. I like what you're going for to avoid a breaking major change by introducing a new lifecycle instead of changing up `webpack`. What about a different name, maybe just `webpackConfig`?

Also, the last argument would be most flexible as an object, in case we want to pass along other helpers down the road. For example, it would be really good if we could pass down the imported Webpack version, too.

`execWaterfallSync('webpackConfig', config, data, { webpack, webpackMerge })`

DullReferenceException: Great ideas.

DullReferenceException: @kinetifex: updated, renaming to `webpackConfig` passing an options object.

packages/gasket-plugin-webpack/index.js
```diff
@@ -6,29 +7,36 @@ const { name, devDependencies } = require('./package');
...
 * @param  {Gasket} gasket The Gasket API
-* @param {Object} webpackConfig Initial webpack config
-* @param {Object} data Additional info
+* @param {Object} initConfig Initial webpack config
+* @param {Object} nextOptions Additional info
 * @returns {Object} Final webpack config
 */
-function initWebpack(gasket, webpackConfig, data) {
-  const { execSync, config } = gasket;
+function initWebpack(gasket, initConfig, nextOptions) {
```
kinetifex, line 14: This function could be used by non-next apps to initialize webpack. If you want to change this variable name then `context` would more general and appropriate.
```suggestion
function initWebpack(gasket, initConfig, context) {
```

packages/gasket-plugin-webpack/index.js
```diff
@@ -6,29 +7,36 @@ const { name, devDependencies } = require('./package');
...
+  const mergedConfig = webpackMerge.smart(
+    baseConfig,
+    ...configPartials
+  );
 
-  return webpackMerge.smart(webpackConfig, ...configs);
+  return execWaterfallSync(
+    'webpackConfig',
+    mergedConfig,
+    { webpack, nextOptions, webpackMerge }
```
kinetifex, line 38: This should next be Next-specific. See the previous comment. If we call this the context, instead just add the helpers to it:
```suggestion
    { ...context, webpack, webpackMerge }
```

kinetifex (changes requested):

packages/gasket-plugin-webpack/index.js
jpage-godaddy: Updated.

packages/gasket-plugin-webpack/index.js
jpage-godaddy: Updated

packages/gasket-plugin-webpack/README.md
```diff
@@ -120,6 +124,39 @@ function webpackHook(gasket, config) {
...
+   
+
+A hook should return a new webpack config object derived from the original. The usage of the `webpack-merge` API is recommended when doing so since properly handling the overloaded types within webpack config properties can be tricky. We recommend avoiding `webpack-merge` methods that have been deprecated in version 5 since a future version of this plugin may update to a new breaking version of `webpack-merge`.
+
+
+```js
+function webpackHook(
+  gasket,
+  config,
+  { context: { isServer }, webpack, webpackMerge }
```
kinetifex, line 146: Small doc name nit and fix to context sig:

```suggestion
function webpackConfigHook(
  gasket,
  config,
  { isServer, webpack, webpackMerge }
```

kinetifex (approved):

burgessa23 (approved):

PR #287, opened 2021-07-07
Add HTTP/2 support

kinetifex:
## Summary

This works with the changes from https://github.com/http-party/create-servers/pull/37 to allow Gasket apps to work configure servers using HTTP/2. The h2 servers are created with the [compatibility API](https://nodejs.org/api/http2.html#http2_compatibility_api). This works fine for Fastify, however, for current Express, some bridging was necessary to put in place.

## Changelog

**@gasket/plugin-https**
- Add support for http2 servers

**@gasket/plugin-express**
- Add support for using http2 servers

## Test Plan

- local app testing
- expanded unit tests

### DIFF
--- a/packages/gasket-plugin-express/index.js
+++ b/packages/gasket-plugin-express/index.js
@@ -20,7 +20,7 @@ module.exports = {
       });
 
       if (context.apiApp) {
-        context.files.add(`${__dirname}/generator/**/*`);
+        context.files.add(`${ __dirname }/generator/**/*`);
 
         context.gasketConfig.add('express', {
           routes: './routes/*'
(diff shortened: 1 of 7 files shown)


packages/gasket-plugin-express/index.js
```diff
@@ -42,9 +42,26 @@ module.exports = {
       const compression = require('compression');
 
       const { config } = gasket;
-      const { root, express: { routes } = {} } = config;
+      const { root, express: { routes } = {}, http2 } = config;
       const excludedRoutesRegex = config.express && config.express.excludedRoutesRegex;
-      const app = express();
+      const app = http2 ? require('http2-express-bridge')(express) : express();
```
mswaagman-godaddy, line 47: Do we really need this dependency? It seems to be just providing fallback to http1 if alpn is not supported from the middleware https://github.com/rahulramesha/http2-express-bridge/blob/main/src/initMiddleware.js#L15-L16 but otherwise it does a lot of prototype mixing and merging and setting some defaults for the router.

Isn't all of this encapsulated by the `allowHTTP1` option in https://nodejs.org/api/http2.html#http2_http2_createsecureserver_options_onrequesthandler

mswaagman-godaddy (commented):
Overall looking good, but got some reservations wrt the use of http2-express-bridge

packages/gasket-plugin-express/index.js
kinetifex: It is a bit more than just that. [This article](https://javascript.plainenglish.io/serving-hello-world-with-http2-and-express-js-4dd0ffe76860) explains the problem well, but `tl;dr` is that Express replaces req and res with custom objects that are incompatible with the http2 object types, which is what this is fixing.

mswaagman-godaddy: TIL express does prototype overwrites 😬

mswaagman-godaddy (approved):

jpage-godaddy (approved):

PR #289, opened 2021-07-28
Updated the default .babelrc to use only set modules: 'commonjs' for the test env.

kawikabader:
## Summary

We recently discovered that having `modules: 'commonjs'` in the .babelrc interfere with next/dynamic chunk determination. This was only in place for test frameworks.

## Changelog

Updated the default .babelrc to use only set modules: 'commonjs' for the test env.

## Test Plan

Manually tested and all existing unit tests are passing.


smorrissey1-godaddy (approved):

burgessa23 (approved):

rmarkins-godaddy (approved):

packages/gasket-plugin-nextjs/generator/app/.babelrc
```diff
@@ -1,12 +1,16 @@
 {
-  "presets": [
```
kinetifex, line 2: I tried this against the canary app. While tests work (`npm test`), builds (`npm run build`) break with it.

It looks like having `next/babel` without customization in the base preset definition works:
```
  "presets": [
    "next/babel"
  ],
  "env:" ...
```

kinetifex (changes requested):
Vet this against the canary-app and others to make sure test + builds continue to work.

kinetifex (approved):

PR #300, opened 2021-11-04
Solving reference binding of module imports bug and stubs during testing

mkasper1-godaddy:
## Summary

Babel converts exports * imports to getters to correctly implement reference binding of module imports. This causes issues when needing to stub out one of the exports during testing. If a consuming app tries to stub out one of the exports, the stub dosent work.

This PR solves that issue by reworking how the components are exported. This change allows the stubs to function properly.

## Changelog

Changed the index.js so it dosent export * import, but instead exports consts

## Test Plan

no change needed

### DIFF
--- a/packages/gasket-nextjs/package.json
+++ b/packages/gasket-nextjs/package.json
@@ -13,7 +13,7 @@
     "test:watch": "npm run test:runner -- --watch",
     "test:coverage": "nyc --reporter=text --reporter=json-summary npm run test:runner",
     "posttest": "npm run lint",
-    "build": "babel src -d lib",
+    "build": "rimraf lib && babel src -d lib",
     "prepublishOnly": "npm run build"
   },
   "repository": {
@@ -65,7 +65,8 @@
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "setup-env": "^1.2.2",
-    "sinon": "^9.2.3"
+    "sinon": "^9.2.3",
+    "rimraf": "^3.0.2"
   },
   "peerDependencies": {
     "next": "^10.2.0"
--- a/packages/gasket-nextjs/src/index.js
+++ b/packages/gasket-nextjs/src/index.js
@@ -2,6 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Main, NextScript } from 'next/document';
 import htmlescape from 'htmlescape';
+import { useGasketData as _useGasketData } from './use-gasket-data';
+import { withGasketDataProvider as _withGasketDataProvider } from './with-gasket-data-provider';
+
 
 /**
  * Renders a script tag with JSON gasketData
@@ -95,5 +98,6 @@ export function withGasketData(options = {}) {
   };
 }
 
-export * from './use-gasket-data';
-export * from './with-gasket-data-provider';
+
+export const useGasketData = _useGasketData;
+export const withGasketDataProvider = _withGasketDataProvider;


packages/gasket-nextjs/src/index.js
```diff
@@ -2,6 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Main, NextScript } from 'next/document';
 import htmlescape from 'htmlescape';
+import { useGasketData as _useGasketData } from './use-gasket-data';
+import { withGasketDataProvider as _withGasketDataProvider } from './with-gasket-data-provider';
```
kinetifex, line 6: If these are written more consicely using export from, does that solve your use case?
```suggestion
export { useGasketData } from './use-gasket-data';
export { withGasketDataProvider } from './with-gasket-data-provider';
```

kinetifex (commented):
Please rebase this against the master branch.

packages/gasket-nextjs/src/index.js
mkasper1-godaddy: unfortunately it dosent. With this approach babel does the same thing as the original way i had it. it converts the exports to getter functions and those dont place nice with stubbing

packages/gasket-nextjs/package.json
```diff
@@ -13,7 +13,7 @@
     "test:watch": "npm run test:runner -- --watch",
     "test:coverage": "nyc --reporter=text --reporter=json-summary npm run test:runner",
     "posttest": "npm run lint",
-    "build": "babel src -d lib",
+    "build": "rimraf lib && babel src -d lib",
```
mkasper1-godaddy, line 16: i noticed the deployed package had some old artifacts from prior iterations of the plugin. Added this to remove the old ones before deployment.

jpage-godaddy:
Can you provide an example of some stubbing that doesn't work? I'd rather we fix your stubbing approach instead of doing exports in a strange way to accommodate.

mkasper1-godaddy:
Here is how im stubbing:

```
import * as nextJs from "@gasket/nextjs";
sinon.stub(nextJs, "useGasketData").returns({
    routes: [{
        name: 'test',
        path: '/test',
        page: 'index'
    }]
});
```

This seems to be the recommended way. I have also tried `sinon.spy` and `proxyquire`. Neither work. Open to suggestions.

jpage-godaddy:
`proxyquire` should work, though it can be a pain since it only mocks direct dependencies. It's unfortunate you're not using jest.

If you want to replace an existing function with a stub you could always work around with [`Object.defineProperty`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty); I've used that to replace getters successfully in the past:

```javascript
Object.defineProperty(nextJs, "useGasketData", {
  value: sinon.stub().returns({
    routes: [{
      name: 'test',
      path: '/test',
      page: 'index'
    }]
});
```

...however this will not help you if something has already imported and destructured out the original property value.

mkasper1-godaddy:
I agree on the Jest thing. The only reason im using sinon and assume is it seemed to be the standard way other gasket plugins are testing and i was trying to keep things consistent. If thats not a thing then ill switch to jest. Dont think i have ever bumped into this issue with jest. I do get nerves about others bumping into this issue. Im able to stub other packages great. Really odd thing.

jpage-godaddy:
Oh, tip on `proxyquire`, I find I pretty much always have to use `proxyquire.noCallThru()` when stubbing imports.

kinetifex:
Yea, it's a mixed bag of jest vs mocha + sinon, etc. right now. Either way, this should work with any test suite. I think @jpage-godaddy is correct that the noCallThru() is probably what is actually tripping you up.

mkasper1-godaddy:
was able to get things working with `noCallThru()` thanks

PR #301, opened 2021-11-11
[Bug] Lifecycle Plugin Main Field

kawikabader:
## Summary

Started receiving `[DEP0128] DeprecationWarning: Invalid 'main' field` for the lifecycle plugin. Removed incorrect main property in package.json.

## Changelog

Removed incorrect main property in package.json.

### DIFF
--- a/packages/gasket-plugin-lifecycle/package.json
+++ b/packages/gasket-plugin-lifecycle/package.json
@@ -2,7 +2,6 @@
   "name": "@gasket/plugin-lifecycle",
   "version": "6.7.3",
   "description": "Allows a gasket/ directory to be used for lifecycle hooks in applications.",
-  "main": "lib/index.js",
   "scripts": {
     "lint": "eslint *.js test/*.js",
     "lint:fix": "npm run lint -- --fix",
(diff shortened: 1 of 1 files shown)


runruh-godaddy (approved):

packages/gasket-plugin-lifecycle/package.json
```diff
@@ -2,7 +2,6 @@
   "name": "@gasket/plugin-lifecycle",
   "version": "6.7.3",
   "description": "Allows a gasket/ directory to be used for lifecycle hooks in applications.",
-  "main": "lib/index.js",
```
kinetifex, line 5: Maybe leave the `main` property, but correct the value, as we have in other plugins:
```
  "main": "index.js"
```

Though this is the default according to [the docs](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#main).

kinetifex (approved):

jpage-godaddy (approved):

packages/gasket-plugin-lifecycle/package.json
kawikabader: Correct, the default is `index.js`, which was why I removed the property entirely, but happy to add it back if it adds clarity.

burgessa23 (approved):

PR #311, opened 2021-12-10
TypeScript definitions for config & lifecycles

jpage-godaddy:
## Summary

- Create types for the core gasket engine
- Create core interfaces which plugins can extend
- Add config & lifecycle typings for all plugins
- Add unit tests to check that type validation works

## Changelog

TBD


packages/gasket-plugin-https/index.d.ts
```diff
@@ -0,0 +1,71 @@
...
+    root?: string,
+    key: MaybeMultiple<CertInput>,
+    cert: MaybeMultiple<CertInput>,
+    ca?: MaybeMultiple<CertInput>,
+    ciphers?: MaybeMultiple<string>,
+    honorCipherOrder?: boolean
+  };
+
+  type HttpsSettings = 
+    & CustomHttpsSettings
```
decompil3d, line 27: Should there be something before the `&` or is this a typescript syntax that I'm not familiar with?

packages/gasket-plugin-service-worker/package.json
```diff
@@ -33,13 +34,16 @@
   },
   "homepage": "https://github.com/godaddy/gasket/tree/master/packages/gasket-plugin-service-worker#readme",
   "dependencies": {
+    "@types/lru-cache": "^5.1.1",
+    "@types/uglify-js": "^3.13.1",
     "deepmerge": "^4.2.2",
     "lru-cache": "^5.1.1",
     "mkdirp": "^0.5.1",
     "uglify-js": "^3.5.3",
```
decompil3d, line 42: Not really related to this PR, but @kinetifex -- should this move to `terser`?

decompil3d (approved):

packages/gasket-plugin-https/index.d.ts
jpage-godaddy: It's something TypeScript supports. `|` as well. It's just a syntax thing that lets you do a list of unions/intersections in a nicer-looking way.

decompil3d: Ok, that's fine. I'm just surprised that a leading `&` is allowed. I'd have expected something more like:

```ts
type HttpsSettings =
  CustomHttpsSettings &
  Omit<
     SecureContextOptions,
     keyof CustomHttpsSettings | 'secureProtocol' | 'secureOptions'
     >;
```

packages/gasket-plugin-service-worker/package.json
kinetifex: It should be updated, but we'll probably go with [swc modification](https://swc.rs/docs/configuration/minification).

packages/gasket-plugin-config/lib/plugin.d.ts
```diff
@@ -0,0 +1,17 @@
+import type { GasketConfig, HookExecTypes, MaybeAsync } from '@gasket/engine';
```
kinetifex, line 1: Several of these `.d.ts` files have unused imports for GasketConfig and/or HookExecTypes. My IDE picks up on it, but is that, expected? Should we go ahead and introduce eslint settings for `.ts`?

packages/gasket-plugin-https/package.json
```diff
@@ -3,6 +3,11 @@
   "version": "6.5.1",
   "description": "Create http/s servers with graceful termination",
   "main": "index.js",
+  "types": "index.d.ts",
+  "files": [
+    "index.js",
+    "index.d.ts"
+  ],
```
kinetifex, line 10: Just wondering, why is this the only package.json that needs the `files` property?

kinetifex (changes requested):
Very cool and thorough @jpage-godaddy! There are some lifecycles that are executed from the `@gasket/cli` package that were missed. Also, what are you thinking for the non-plugin packages? - Not going to block on those for the PR, but it'd be good to have follow-up tickets/issues so we do not lose track.

The only thing blocking really is just marking the test package as private. Besides that, some possible lint cleanup, and what are your thoughts on a follow-up to include `.ts` type files with our linting rules?

packages/gasket-plugin-https/index.d.ts
kinetifex: Perhaps we can lint on code styles here ([see comment](https://github.com/godaddy/gasket/pull/311#discussion_r769998926)).

packages/gasket-plugin-config/lib/plugin.d.ts
jpage-godaddy: I was under the impression that you had to import them before you could extend but I'm not 100% sure on that. Let me review and see if I can remove the unused imports.

jpage-godaddy: Looks like those are not necessary after all; removed.

packages/gasket-plugin-https/package.json
jpage-godaddy: I added here because this package is missing an `.npmignore` so it's currently publishing unnecessary files. Added `files` in here to fix rather than `.npmignore` since IMO that's a far superior approach that requires less upkeep.

kinetifex: Got it - yes `files` is my preferred approach as well.

packages/gasket-engine/package.json
```diff
@@ -46,7 +46,7 @@
     "eslint-plugin-jest": "^24.1.3",
     "eslint-plugin-json": "^2.1.2",
     "eslint-plugin-mocha": "^8.0.0",
-    "jest": "^26.6.3"
+    "jest": "^27.4.3"
```
kinetifex, line 49: I was investigating the failing tests and came across this as well. To pick up all packages, please update the align-packages script [here](https://github.com/jpage-godaddy/gasket/blob/type-decls/scripts/align-packages.js#L37). Then run `npm run align-packages` at the root of the repo.

Before you run tests, you also need to update this particular spec file:
`packages/gasket-redux/test/configure-make-store.spec.js`
to include the following at the top ([docs](https://jestjs.io/docs/configuration#testenvironment-string)):
```
/**
 * @jest-environment jsdom
 */
```

jpage-godaddy: Thanks, that fixed it.

jpage-godaddy:
> There are some lifecycles that are executed from the `@gasket/cli` package that were missed.

Thanks. Added those.

> Also, what are you thinking for the non-plugin packages?

It can be much the same, just adding to a `.d.ts` (or converting the package to TypeScript and auto-generating that, which is easier IMO). Could be pretty incremental, maybe could have a ticket per package?

> The only thing blocking really is just marking the test package as private.

Fixed.

> ...what are your thoughts on a follow-up to include `.ts` type files with our linting rules?

If eslint supports that sure! I'd think we'd have to switch to `eslint-config-godaddy-typescript` or similar in order for TypeScript to be recognized, but I don't know how well that works when the majority of the codebase is not TypeScript.

kinetifex (approved):

rxmarbles (approved):

PR #315, opened 2022-01-10
Added healthcheck.html to terminus options.

kawikabader:
## Summary

There is a need to support existing apps that are using the `/healthcheck.html` route instead of the `/healthcheck`.

## Changelog

Added the `/healthcheck.html` route to the Terminus `healthChecks` options.

## Test Plan

Tests were updated.

### DIFF
--- a/packages/gasket-plugin-https/index.js
+++ b/packages/gasket-plugin-https/index.js
@@ -37,13 +37,13 @@ async function start(gasket) {
   if (http2) configOpts.http2 = http2;
 
   const serverOpts = await gasket.execWaterfall('createServers', configOpts);
-  const { healthcheck, healthcheckHtmlRoute, ...terminusDefaults } = await gasket.execWaterfall('terminus', {
-    healthcheck: '/healthcheck',
-    healthcheckHtmlRoute: '/healthcheck.html',
+  const { healthcheck, ...terminusDefaults } = await gasket.execWaterfall('terminus', {
+    healthcheck: ['/healthcheck', '/healthcheck.html'],
     signals: ['SIGTERM'],
-
     ...(terminus || {})
   });
+
+  const routes = Array.isArray(healthcheck) ? healthcheck : [healthcheck];
 
   // Default port to non-essential port on creation
   // create-servers does not support http or https being `null`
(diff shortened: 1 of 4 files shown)


packages/gasket-plugin-https/index.js
```diff
@@ -37,8 +37,9 @@ async function start(gasket) {
   if (http2) configOpts.http2 = http2;
 
   const serverOpts = await gasket.execWaterfall('createServers', configOpts);
-  const { healthcheck, ...terminusDefaults } = await gasket.execWaterfall('terminus', {
+  const { healthcheck, healthcheckHtmlRoute, ...terminusDefaults } = await gasket.execWaterfall('terminus', {
     healthcheck: '/healthcheck',
+    healthcheckHtmlRoute: '/healthcheck.html',
```
kinetifex, line 42: Glad we can allow for multiple routes. To simplify the config signature, and keep it flexible, could we default `healthcheck` to an array of route names, but which supports a single string if set?
```
const routes = Array.isArray(healthcheck) ? healthcheck : [healthcheck]
```

runruh-godaddy: > Glad we can allow for multiple routes. To simplify the config signature, and keep it flexible, could we default `healthcheck` to an array of route names, but which supports a single string if set?
> 
> ```
> const routes = Array.isArray(healthcheck) ? healthcheck : [healthcheck]
> ```

I love this idea as well

packages/gasket-plugin-https/index.js
```diff
@@ -74,12 +75,15 @@ async function start(gasket) {
     onShutdown: one(async function onShutdown() {
       await gasket.exec('onShutdown');
     }),
-    healthChecks: {
-      [healthcheck]: async function healthCheckRequested() {
-        await gasket.exec('healthcheck', HealthCheckError);
+    healthChecks: (function () {
```
kawikabader, line 78: @kinetifex @runruh-godaddy would this be more aligned with what you were looking for?

kinetifex: Yea, avoid the IIFE though for readability, just use a reduce:
```js
async function healthCheckRequested() {
  await gasket.exec('healthcheck', HealthCheckError);
}

healthChecks: routes.reduce((acc, cur) => {
  acc[cur] = healthCheckRequested;
  return acc;
}, {})
```

burgessa23 (approved):

kinetifex:
Looks good and I tested it successfully in the canary-app. Just update the configuration docs with the new defaults then 👍🏻

kinetifex (approved):

Issue #314, opened 2022-01-07
Use promise modules


kinetifex:
## Description
We use `promisify` for several things that now already have promise support. Let's do a code cleanup pass for these.

- `fs` -> `fs.promises`
- `mkdirp@0.5.1` -> `mkdirp@^1.x`

## AC:
* Do not use promisify with fs (use fs.promises)
* Align `mkdirp` dependency across packages
* Check for updates to other packages and avoid `promisify` where possible

PR #319, opened 2022-01-14
Convert to `fs.promises` and Align `mkdirp`

rxmarbles:
## Summary

This PR solves #314. I've aligned `mkdirp` in all packages (which this version has native promise support). Along w/ moving to `fs.promises` vs `promisify` for all things `fs` related

## Changelog

- minor: convert to `require('fs').promises` and align `mdirp` accross packages

## Test Plan

Updated all tests related to the `require('fs').promises` test suite passes locally

### DIFF
--- a/packages/gasket-cli/src/scaffold/fetcher.js
+++ b/packages/gasket-cli/src/scaffold/fetcher.js
@@ -168,9 +168,10 @@ module.exports = class PackageFetcher {
    */
   async unpack(tarball, dir) {
     return {
-      then: (fulfill, reject) => {
+      then: async (fulfill, reject) => {
         const logOpts = { tarball, dir };
-        const fd = open(tarball);
+        const fd = await open(tarball);
+        console.log(fd);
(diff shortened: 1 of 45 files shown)


jpage-godaddy:
Will this obsolete node 12 support? To use the promises API in node 12 you have to use a pattern like this instead:

```javascript
const { readFile } = require('fs').promises;
```

jpage-godaddy (approved):

packages/gasket-cli/src/scaffold/fetcher.js
```diff
@@ -174,7 +170,8 @@ module.exports = class PackageFetcher {
     return {
       then: (fulfill, reject) => {
         const logOpts = { tarball, dir };
-        const readableStream = fs.createReadStream(tarball).once('error', this._logError(`fs.createReadStream`, logOpts));
+        const fd = open(tarball);
```
kinetifex, line 173: Don't you need to await this?
```suggestion
        const fd = await open(tarball);
```

kinetifex (approved):

kawikabader (approved):

PR #322, opened 2022-01-21
Fix CI Duplicate Running

kawikabader:
## Summary

Currently, when updates are pushed to a branch that is in PR, the CI jobs are duplicated.

image

## Changelog

Update the CI workflow to only run when a PR is opened, reopened, or taken out of draft mode. Also, continue to run on push.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,6 @@ name: CI
 
 on:
   pull_request:
-    types: [opened, reopened, ready_for_review]
   push:
     branches:
       - 'main'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,11 +2,13 @@
 
 name: CI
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -2,11 +2,13 @@
 
 name: CI
 
-on: [push, pull_request]
+on:
+  pull_request:
+    types: [opened, reopened, ready_for_review]
+  push:
```
rxmarbles, line 8: this will still trigger multiple because you are not specifing a branch.  From what I recall if you do 

```suggestion
  push:
    branches:
      - 'main'
```

that will still trigger. the ci will still trigger on each push to a PR because it is opened and ready for review

rxmarbles (changes requested):

kinetifex (approved):

rxmarbles (approved):

PR #346, opened 2022-03-10
Deprecating webpack-merge smart approaches

kinetifex:
## Summary

We were relying on the "smart merge" method from [webpack-merge] which was deprecated in v4 and has been removed in v5 ([see changelog]). Further, we had several approaches for configuring custom Webpack which was sometimes confusing.

This PR sets things up to remove Webpack merging from being handled by the Gasket Webpack plugin. This will put control in apps and plugins to merge and open up more recent version of [webpack-merge].

## Changelog

**@gasket/plugin-webpack**
- Log deprecated warnings and add docs pointing to use `webpackConfig` lifecycle

## Test Plan

- Local testing
- Updated unit tests

[webpack-merge]: https://github.com/survivejs/webpack-merge
[see changelog]: https://github.com/survivejs/webpack-merge/blob/master/CHANGELOG.md#503--2020-07-06

### DIFF
--- a/packages/gasket-plugin-webpack/README.md
+++ b/packages/gasket-plugin-webpack/README.md
@@ -30,7 +30,7 @@ module.exports = {
 
 ## Configuration
 
-The webpack plugin is configured using the `gasket.config.js` file.
+The Webpack plugin is configured using the `gasket.config.js` file.
 
 First, add it to the `plugins` section of your `gasket.config.js`:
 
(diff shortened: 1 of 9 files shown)


packages/gasket-plugin-webpack/README.md
```diff
@@ -85,12 +70,51 @@ const config = initWebpack(gasket, webpackConfig, data);
 
 ## Lifecycles
 
+### webpackConfig
+
+Executed with initial, it receives three parameters:
```
jpage-godaddy, line 75: Not clear what "Executed with initial" means. Is it supposed to be `Executed by `initWebpack`?"

jpage-godaddy (approved):

indexzero (approved):
✅ solid upgrade for the functionality that we originally used from webpack-merge. iirc original implementation was from me way back when so gave it a review 🫡

> 📜 Tangential area that may be of interest for related inspiration: pattern gasket leveraged was inspired (and attributed in some places under MIT) to vue-cli. If the feature is deprecated in webpack-merge itself might be interesting to read what they do with the latest version

burgessa23 (approved):

PR #359, opened 2022-05-13
@gasket/plugin-config_error_handling

mmason2-godaddy:
## Summary
Descriptive error handling when hooking into the `appRequestConfig` lifecycles. Currently if a config object is not returned from the handlers it results in an error [here](https://github.com/godaddy/gasket/blob/main/packages/gasket-plugin-config/lib/middleware.js#L20).

Added descriptive error handling for `appEnvConfig` lifecycle [here](https://github.com/godaddy/gasket/blob/main/packages/gasket-plugin-config/lib/plugin.js#L12-L15)

## Changelog
Unclear if this change should be added to the change log.

## Test Plan
`appRequestConfig` - `middleware.test.js` - Added test case to ensure that an error is thrown and has a matching description.
`appEnvConfig` - `plugin.test.js` - Added test case to ensure that an error is thrown and has a matching description.

### DIFF
--- a/packages/gasket-plugin-config/lib/middleware.js
+++ b/packages/gasket-plugin-config/lib/middleware.js
@@ -16,6 +16,14 @@ function handler(gasket) {
         req,
         res);
 
+      // Error handling
+      // Better error description for the appRequestConfig & appEnvConfig lifecycles
+      // In the event a config object isn't returned from the hook(s)
+      // it will result in a error in the destructuring of req.config
+      if (!req.config) {
+        throw new Error('req.config is undefined - Ensure that all plugins hooking into the appRequestConfig and appEnvConfig lifecycle return a config object.')
+      }
+
       const { gasketData = {} } = res.locals;
       const { public: config = {} } = req.config;
 
--- a/packages/gasket-plugin-config/lib/middleware.js
+++ b/packages/gasket-plugin-config/lib/middleware.js
@@ -21,7 +21,10 @@ function handler(gasket) {
       // In the event a config object isn't returned from the hook(s)
       // it will result in a error in the destructuring of req.config
       if (!req.config) {
-        throw new Error('req.config is undefined - Ensure that all plugins hooking into the appRequestConfig and appEnvConfig lifecycle return a config object.')
+        throw new Error(
+          // eslint-disable-next-line max-len
+          'req.config is undefined - Ensure that all plugins hooking into the appRequestConfig and appEnvConfig lifecycle return a config object.'
+        );
       }
 
       const { gasketData = {} } = res.locals;
--- a/packages/gasket-plugin-config/lib/middleware.js
+++ b/packages/gasket-plugin-config/lib/middleware.js
@@ -17,13 +17,12 @@ function handler(gasket) {
         res);
 
       // Error handling
-      // Better error description for the appRequestConfig & appEnvConfig lifecycles
+      // Better error description for the appRequestConfig lifecycles
       // In the event a config object isn't returned from the hook(s)
       // it will result in a error in the destructuring of req.config
       if (!req.config) {
         throw new Error(
-          // eslint-disable-next-line max-len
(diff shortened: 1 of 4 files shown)


packages/gasket-plugin-config/lib/middleware.js
```diff
@@ -16,6 +16,17 @@ function handler(gasket) {
...
         res);
 
+      // Error handling
+      // Better error description for the appRequestConfig & appEnvConfig lifecycles
+      // In the event a config object isn't returned from the hook(s)
+      // it will result in a error in the destructuring of req.config
+      if (!req.config) {
+        throw new Error(
+          // eslint-disable-next-line max-len
+          'req.config is undefined - Ensure that all plugins hooking into the appRequestConfig and appEnvConfig lifecycle return a config object.'
```
kinetifex, line 26: 👍🏻 Let's make the error message concise and specific. Can you do a check and throw for `appEnvConfig` [here](https://github.com/godaddy/gasket/blob/main/packages/gasket-plugin-config/lib/plugin.js#L12-L15).

You messaging could then be something like:
`An appRequestConfig lifecycle hook did not return a config object.`
and
`An appEnvConfig lifecycle hook did not return a config object.`

kinetifex (approved):
Great investigation and contribution! I left a couple of small cleaning comments - then 👍🏻

packages/gasket-plugin-config/test/plugin.test.js
```diff
@@ -155,6 +155,23 @@ describe('Plugin', () => {
...
+      Object.assign(gasket.config, {
+        root: path.join(__dirname, './fixtures', 'templated'),
+        env: 'test'
+      });
+
+      gasket.execWaterfall.mockImplementation((event) => {
+        expect(event).toEqual('appEnvConfig');
+        // eslint-disable-next-line no-undefined
+        return undefined;
+      });
```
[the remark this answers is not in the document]
mmason2-godaddy: Yeah this was definitely a brain lapse on my part 🤦‍♂️ 

Great catch!

packages/gasket-plugin-config/lib/middleware.js
```diff
@@ -16,6 +16,16 @@ function handler(gasket) {
         req,
         res);
 
+      // Error handling
+      // Better error description for the appRequestConfig lifecycles
+      // In the event a config object isn't returned from the hook(s)
+      // it will result in a error in the destructuring of req.config
```
kawikabader, line 22: Is this comment necessary?

packages/gasket-plugin-config/lib/plugin.js
```diff
@@ -13,6 +13,13 @@ module.exports = {
         'appEnvConfig',
         mergeRootConfig(gasket, mergeConfigFiles(gasket))
       );
+
+      // Better error description for the appRequestConfig lifecycles
```
kawikabader, line 17: Do we need this comment?

packages/gasket-plugin-config/lib/middleware.js
```diff
@@ -16,6 +16,16 @@ function handler(gasket) {
...
+      if (!req.config) {
+        throw new Error(
```
kawikabader, line 24: Should we use the `gasket.logger` object here?

packages/gasket-plugin-config/lib/plugin.js
```diff
@@ -13,6 +13,13 @@ module.exports = {
...
+      if (!gasket[ENV_CONFIG]) {
+        throw new Error(
```
kawikabader, line 19: Should we use the `gasket.logger` object here?

kawikabader (approved):
Just a couple of questions.

packages/gasket-plugin-config/lib/middleware.js
kinetifex: I think a hard throw on these here makes sense. It should fail the app.

packages/gasket-plugin-config/lib/middleware.js
mmason2-godaddy: It's verbose and could be removed.

PR #364, opened 2022-05-27
PFX-236 - Add support for configuration of the Docusaurus docs directory name

mmason2-godaddy:
## Summary

Add support for configuration of the Docusaurus docs directory name.

[PFX-236](https://jira.godaddy.com/browse/PFX-236)

## Changelog

Add support for configuration of the Docusaurus docs directory name.

## Test Plan
Tested changes with canary publish. Ran "gasket create" and it executed without issue.

[6.23.0-canary.0](https://www.npmjs.com/package/@gasket/plugin-docusaurus/v/6.23.0-canary.0)

Screen Shot 2022-05-27 at 4 16 13 PM

### DIFF
--- a/packages/gasket-plugin-docusaurus/README.md
+++ b/packages/gasket-plugin-docusaurus/README.md
@@ -36,7 +36,8 @@ launched in your browser.
 
 To be set under `docusaurus` in the `gasket.config.js`.
 
-- `docsRoot` - (string) Name of the root documents folder. Default is `.docs`.
+- `rootDir` - (string) Root Docusaurus directory. Default is `.docs`.
+- `docsDir` - (string) Sub-directory for the generated markdown from the docs plugin. Default is `docs`.
 - `port` - (number) Port to serve the docs from. Default is `3000`.
 - `host` - (string) Hostname to serve the docs from. Default is `localhost`.
 
@@ -45,21 +46,19 @@ To be set under `docusaurus` in the `gasket.config.js`.
 // gasket.config.js
 
 module.exports = {
-  docs: {
-    outputDir: 'documentation' // default is "docs"
-  },
   docusaurus: {
-    docsRoot: 'my-site-documents',
+    rootDir: 'my-site-documents',
+    docsDir: 'markdown',
     port: 8000,
     host: 'custom-host'
   }
 };
 
-// directories
+// structure
 gasket-app/ // app root
 |_ my-site-documents // docusaurus root
   |_ .docusaurus // build folder
-  |_ documention // generated docs from @gasket/plugin-docs
+  |_ markdown // generated docs from @gasket/plugin-docs
 ```
 
 ### `docusaurus.config.js`
--- a/packages/gasket-plugin-docusaurus/README.md
+++ b/packages/gasket-plugin-docusaurus/README.md
@@ -36,7 +36,7 @@ launched in your browser.
 
 To be set under `docusaurus` in the `gasket.config.js`.
 
-- `docsDir` - (string) Name of the root documents folder. Default is `.docs`.
+- `docsRoot` - (string) Name of the root documents folder. Default is `.docs`.
 - `port` - (number) Port to serve the docs from. Default is `3000`.
 - `host` - (string) Hostname to serve the docs from. Default is `localhost`.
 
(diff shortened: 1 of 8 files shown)


packages/gasket-plugin-docusaurus/lib/configure.js
```diff
@@ -2,20 +2,23 @@ const path = require('path');
 const timing = { before: ['@gasket/plugin-docs'] };
 
 async function handler(gasket, config) {
-  const { docusaurus = {}, docs = {} } = config;
-  const { docsRoot = '.docs' } = docusaurus;
-  const { outputDir = 'docs' } = docs;
+  const { docusaurus = {} } = config;
+  const { rootDir = '.docs', docsDir = 'docs' } = docusaurus;
+
+  if (config.docs && config.docs.outputDir) {
```
mmason2-godaddy, line 8: @kinetifex Moved the warning to the configure lifecycle. On line 16 the `outputDir` is set so having the check for this config property in a later lifecycle would be useless.

packages/gasket-plugin-docusaurus/lib/generate-default-config.js
```diff
@@ -5,10 +5,17 @@ const defaultConfig = path.join(__dirname, '..', 'generator', 'docusaurus.config
...
-module.exports = async function generateDefaultConfig(name) {
-  const configStr = await readFile(defaultConfig, 'utf-8');
-  return configStr.replace(/(\$\{name\})/g, name);
+module.exports = async function generateDefaultConfig(configOptions) {
+  const configProperties = Object.keys(configOptions);
+  let configStr = await readFile(defaultConfig, 'utf-8');
+
+  configProperties.forEach(property => {
+    const regex = `\${${property}}`.replace(/[${}]/g, '\\$&');
+    configStr = configStr.replace(new RegExp(regex, 'g'), configOptions[property]);
```
agerard-godaddy, line 17: I'm not sure exactly what line 16 is supposed to accomplish, but I think I know what you're going for. However, instead of iterating over the config properties, it would be simpler code to iterate of the matches with a [replacer function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_function_as_a_parameter), _then_ get the corresponding config option.

Something like:
```js
const configStr = await readFile(defaultConfig, 'utf-8');
return configStr.replace(/\$\{(\w+)\}/g, (full, propName) => {
    return configOptions[propName];
})
```

agerard-godaddy (approved):

packages/gasket-plugin-docusaurus/README.md
```diff
@@ -36,7 +36,8 @@ launched in your browser.
 
 To be set under `docusaurus` in the `gasket.config.js`.
 
-- `docsDir` - (string) Name of the root documents folder. Default is `.docs`.
+- `rootDir` - (string) Root Docusaurus directory. Default is `.docs`.
+- `docsDir` - (string) Sub-directory for the generated markdown from the docs plugin. Default is `docs`.
```
jpage-godaddy, line 40: I find these names/definitions hard to follow. What is meant by the root Docusaurus directory? Does that mean the root directory for the documentation website? Is `docsDir` a subdirectory within `rootDir`? Do we need both? I can see why someone might want to customize `rootDir` if that is the root of the generated files, but why would someone need to customize `docsDir`?

mmason2-godaddy: -  `rootDir` - The root directory of the Docusaurus application which is basically a React application. Yes it's the root directory for the documentation website.
- `docsDir` - The sub directory for the documentation files within the website root - `.md` `.mdx`.

The `rootDir` isn't the _root of the generated files_. The `docsDir` property is essentially the `docsRoot` for the `@gasket/plugin-docs` and that generated directory needs to be inside the Docusaurus website root.

kinetifex: Let me just add that both dir names can be optionally changed, and are just made available for custom output locations (probably edge-cases). We've gone back and forth on the names for these. And like @mmason2-godaddy mentioned, although they work together, they serve two purposes: `rootDir` - where does my docusaurus app go, `docsDir` - where does the generated markdown go within my docusaurus app.

jpage-godaddy (approved):

jpage-godaddy:
I'm hoping discussion here can help us come up with a more objective decision framework:

https://github.secureserver.net/CTO/guidelines/pull/594

PR #370, opened 2022-06-06
Support for --require flags

agerard-godaddy:
## Summary

Adding  support for a `--require` (or `-r`) to allow apps to preload modules before the CLI fully initializes. This pattern is in line with other familiar CLIs such as [node](https://nodejs.org/api/cli.html#-r---require-module) or [mocha](https://mochajs.org/#-require-module-r-module).

We are using [getopts](https://github.com/jorgebucaran/getopts) as an intermediate args parser for this step. It is a very thin zero-dependency parser that avoids bringing in several dependencies which developers may desire bootstrapping.

### Background 

Exploring options for loading instrumentation tool for various Gasket commands. In particular, using [@gasket/plugin-elastic-apm](https://github.com/godaddy/gasket/tree/main/packages/gasket-plugin-elastic-apm) to load [apm-agent-nodejs](https://github.com/elastic/apm-agent-nodejs) is too late in the process. While it fires in the `preboot` lifecycle, this happens after certain modules have already been required and so miss being bootstrapped.

## Changelog

**@gasket/cli**
- Support for `--require` flag to load modules before Gasket initializes

## Test Plan

- [x] updated unit test
- [x] locally with `create` command and `start` command in an app

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18569,6 +18569,11 @@
         "url": "https://github.com/sponsors/ljharb"
       }
     },
+    "node_modules/getopts": {
+      "version": "2.3.0",
+      "resolved": "https://registry.npmjs.org/getopts/-/getopts-2.3.0.tgz",
+      "integrity": "sha512-5eDf9fuSXwxBL6q5HX+dhDj+dslFGWzU5thZ9kNKUkcPtaPdatmUFKwHFrLb/uf/WpA4BHET+AX3Scl56cAjpA=="
+    },
     "node_modules/getos": {
       "version": "3.2.1",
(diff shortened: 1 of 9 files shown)


agerard-godaddy:
@jpage-godaddy this is going to be our best option. I will have have a followup PR to the `@gasket/plugin-elastic-apm` for new project scaffolding and deprecating its preboot lifecycle.

jpage-godaddy:
This is nice. It'll also unlock things like `gasket local --require ts-node/register`.

mmason2-godaddy (approved):

PR #375, opened 2022-06-08
feat:(@gasket/plugin-docs) Configuration section in generated docs

mmason2-godaddy:
## Summary
Ticket: [PFX-81](https://jira.godaddy.com/browse/PFX-81)
- Allow for `configurations` in metadata config object
- Add support to `DocsConfigSetBuilder` class
- Expand `addSection` in `generate-index.js` to allow for additional table headers & values
- Add configurations alpha sort to `sorts.js`
- Update associated tests, add new tests for the additional table headers & values and add sorts test

## Changelog
- Allow for `configurations` in metadata config object

## Test Plan
Update associated tests, add new tests for the additional table headers & values and add sorts test.

Screen Shot 2022-06-08 at 11 54 38 AM

Screen Shot 2022-06-08 at 11 54 29 AM
---
## Result
Screen Shot 2022-06-08 at 11 48 56 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,8 +129,8 @@ Available structure
 | [plugins/]         | One-off plugins for apps                               |
 | [public/]          | NextJS static files                                    |
 | [public/locales/]  | Locale JSON files with translation strings             |
-| test/              | Test files                                             |
-| test/              | Test files                                             |
+| [test/][6]         | Test files                                             |
+| [test/][7]         | Test files                                             |
 | [app.config.js]    | App configuration with environment overrides           |
 | [cypress.json]     | Cypress configuration                                  |
 | [gasket.config.js] | Gasket config for an app                               |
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -205,6 +205,15 @@ Supporting modules
 | [@gasket/utils]            | 6.24.0  | Reusable utilities for Gasket internals                                              |
 | [create-gasket-app]        | 6.24.1  | starter pack for creating a gasket app                                               |
 
+## Configurations
+
+Available configuration options in the `gasket.config.js`
+
+| Name             | Description                         | Type   | Default | From                |
```
agerard-godaddy, line 212: I feel like the `From` column crowds the table too much and is unnecessary with the name links. I will defer to other opinions. However, if we do keep the column, it seems like they should link, also.

agerard-godaddy (approved):
This is awesome! Just one point of opinionated feedback - but otherwise looks solid.

README.md
mmason2-godaddy: Removed the `From` column in favor of the link fallbacks option(set to true for the configurations generated portion). Links if not specified will point to the package README.

PR #376, opened 2022-06-08
Fix mermaid diagram generation

jpage-godaddy:
- (fix) Add `deprecated` property to lifecycles rather than annotating the name
- (feature) Ensure `(deprecated)` note is shown in the lifecycle diagram

## Summary

The lifecycle diagram code has a syntax error caused by the "(deprecated)" in the node. Remove from the name of the lifecycle and annotate nodes that are deprecated with custom text.

### DIFF
--- a/packages/gasket-plugin-webpack/lib/index.js
+++ b/packages/gasket-plugin-webpack/lib/index.js
@@ -19,13 +19,15 @@ module.exports = {
           link: 'docs/webpack.md'
         }],
         lifecycles: [{
-          name: 'webpackChain (deprecated)',
+          name: 'webpackChain',
+          deprecated: true,
           method: 'execSync',
           description: 'Setup webpack config by chaining',
           link: 'README.md#webpackChain',
           parent: 'initWebpack'
         }, {
-          name: 'webpack (deprecated)',
+          name: 'webpack',
+          deprecated: true,
           method: 'execSync',
           description: 'Modify webpack config with partials or by mutating',
           link: 'README.md#webpack',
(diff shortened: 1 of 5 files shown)


kbehari-godaddy (approved):

jpage-godaddy:
I can't test this locally because I get a `EMFILE: too many open files` error when running `lerna bootstrap` which I can't figure out how to resolve.

agerard-godaddy:
@jpage-godaddy we switched to npm workspaces, so running `lerna bootstrap` should not be necessary anymore. Try just `npm ci` from the repo root.

We only use lerna for running cross-package commands and for versioning and publishing.

packages/gasket-plugin-webpack/lib/index.js
```diff
@@ -19,13 +19,15 @@ module.exports = {
           link: 'docs/webpack.md'
         }],
         lifecycles: [{
-          name: 'webpackChain (deprecated)',
+          name: 'webpackChain',
+          deprecated: true,
```
agerard-godaddy, line 23: I like this idea. It will affect what is displayed in the generated index table for the `gasket docs`, but that's something we could adjust for in the docs plugin.

agerard-godaddy (approved):

packages/gasket-plugin-webpack/lib/index.js
agerard-godaddy: I'll make a follow up ticket for this ☝️

rmarkins-godaddy (approved):

PR #378, opened 2022-06-09
Prefer --require for starting Elastic APM

agerard-godaddy:
## Summary

The [start recommendations] for the APM agent are to require it as early as possible in your app. For Gasket apps, using the new `--require` from https://github.com/godaddy/gasket/pull/370 can accomplish this.

[start recommendations]:https://www.elastic.co/guide/en/apm/agent/nodejs/master/agent-api.html#apm-start

## Changelog

**@gasket/plugin-elastic-apm**
- create hook to add dep and require script
- only start apm agent if not done with --require

## Test Plan

- Updated/added unit tests
- Local app tested:
Screen Shot 2022-06-09 at 11 50 33 AM

### DIFF
--- a/packages/gasket-plugin-elastic-apm/README.md
+++ b/packages/gasket-plugin-elastic-apm/README.md
@@ -72,6 +72,27 @@ module.exports = {
 };
 ```
 
+### Custom Start Configurations
+
+For scenarios where you need to configure the start options for the APM agent,
+you can do so in a custom setup script and require it instead.
+
+For example, add a `setup.js` script to the root of your app:
+
+```
+// setup.js
+require('elastic-apm-node').start({
+  // any configuration options
+})
+```
+
+Then adjust your start script to require it instead:
+
+```diff
+-   "start": "gasket start --require elastic-apm-node/start",
++   "start": "gasket start --require ./setup.js",
+```
+
 ### Custom Filters
 
 According to the [Elastic APM docs], the _Elastic APM agent for Node.js is a
--- a/packages/gasket-plugin-elastic-apm/README.md
+++ b/packages/gasket-plugin-elastic-apm/README.md
@@ -13,7 +13,7 @@ gasket create <app-name> --plugins @gasket/plugin-elastic-apm
 #### Existing apps
 
 ```
-npm i @gasket/plugin-elastic-apm
+npm install @gasket/plugin-elastic-apm elastic-apm-node
 ```
 
 Modify `plugins` section of your `gasket.config.js`:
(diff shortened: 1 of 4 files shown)


packages/gasket-plugin-elastic-apm/test/create.test.js
```diff
@@ -0,0 +1,3 @@
+describe('create lifecycle', function () {
```
jpage-godaddy, line 1: Empty

jpage-godaddy (approved):

jpage-godaddy (approved):

mmason2-godaddy (approved):

packages/gasket-plugin-elastic-apm/lib/index.js
```diff
@@ -29,20 +32,53 @@ module.exports = {
...
     preboot: {
-      handler: async ({ config }) => {
-        require('elastic-apm-node')
-          .start({
+      handler: async (gasket) => {
+        const { config, logger } = gasket;
+
+        // prefer app-level dependency in case of duplicates
+        const apm = require(
+          require.resolve('elastic-apm-node', { paths: [config.root, __dirname] })
```
kbartholomew-godaddy, line 58: Is this a use case for `peerDependencies` instead?

packages/gasket-plugin-elastic-apm/README.md
```diff
@@ -28,35 +28,35 @@ module.exports = {
...
-    add: ['@gasket/plugin-elastic-apm']
-  },
-  elasticAPM: {
-    secretToken: '****',
-    serverUrl: 'http://localhost:9200'
+```diff
+  "scripts": {
+    "build": "gasket build",
+-   "start": "gasket start",
++   "start": "gasket start --require elastic-apm-node/start",
```
kbartholomew-godaddy, line 37: I'm a little concerned that this completely bypasses the plugin and the Gasket config system, because it diminishes our ability to further modify the agent's configuration when it starts. Could the plugin itself expose a module that can be used with the `--require` syntax? It could then read configuration data from either `gasket.config.js` or environment variables and start the agent.

```suggestion
+   "start": "gasket start --require @gasket/plugin-elastic-apm/start",
```

Also, what happens if the upstream repo moves or renames `elastic-apm-node/start`? By not directly requiring a file within an external package, we can provide a better guarantee that this startup method will continue to work if the upstream changes significantly.

packages/gasket-plugin-elastic-apm/lib/index.js
agerard-godaddy: That is correct. Switching the plugin to dev+peer dependencies would be a breaking change for current users of the plugin. So, we'll add it to new apps during create, but leave as a plugin dep for now until the next major release.

packages/gasket-plugin-elastic-apm/README.md
agerard-godaddy: Yes, unfortunately it does bypass the Gasket engine. But, if you want to ensure the agent bootstraps the many modules brought in from loading plugins to set up the Gasket engine, then it has to start before it.

This scaffolds the minimum configuration to get started. `elastic-apm-node/start` is the recommended API given by the elastic docs, so I wouldn't expect it to move. Still, app teams could choose wrap it and do additional configuration in a custom setup script.

```bash
gasket start --require ./setup.js
```
```js
// setup.js
module.exports = {
  require('elastic-apm-node').start({
    // whatever you want
  })
}
```

agerard-godaddy: I can add this as a note in the readme if you think that'd help ☝️

kbartholomew-godaddy: looks great!

PR #395, opened 2022-07-12
Optimize loaded plugin access

agerard-godaddy:
## Summary

We uncovered that configured plugins loading is being called multiple times. This PR introduces a memoize optimization to avoid any impact from getting the loaded plugins again.

## Changelog

**@gasket/resolve**
- Memoize results from plugin loader

## Test Plan

- Updated unit tests


mmason2-godaddy (approved):

jpage-godaddy (approved):
This works ok, but if there's a cleaner way for the metadata plugin to get invoked where it doesn't have to call this function that'd be even better.

kawikabader (approved):

PR #400, opened 2022-07-21
Transaction naming support

jpage-godaddy:
- Support naming and labeling transactions in ElasticAPM
- Add ability to get next.js route details on the serverside
- Add a `locale` label to transactions
- Label/name transactions with next.js details

## Summary

Elastic APM does its best to name transactions (it hooks into express to try to name transactions based on routes), but for situations like next.js apps where our handler is for all routes it doesn't have a way to select good names at present.

This introduces a lifecycle for naming and labeling transactions so you can see requests categorized and helpful metadata when looking at your application's performance. The next.js plugin hooks these lifecycles and uses the routes manifest to select a better transaction name and assign routing parameters as labels. The intl plugin also adds a label for locale. Other plugins that may have useful transaction details can also hook in if we so desire, and of course application users can do their own customizations.


jpage-godaddy:
`preNextHandling` is an oddball here. It's a remnant from an earlier approach I started working on but then abandoned, but I do think it's useful to have the ability to do things before the next.js handler.

packages/gasket-plugin-nextjs/lib/next-route.js
```diff
@@ -0,0 +1,81 @@
...
+    return null;
+  }
+
+  if (cachedRoutes) {
+    return cachedRoutes;
+  }
+
+  if (!routesPromise) {
+    routesPromise = fs
+      .readFile(path.join(gasket.config.root, './.next/routes-manifest.json'))
```
jpage-godaddy, line 56: There doesn't seem to be any sort of official support within next.js to tap into its router that I could find. That would be preferable to reading a manifest JSON file which we don't necessarily have guarantees on for format, but it's the cleanest I could come up with so far. I wish they supported more abilities for custom servers, but I'm guessing they're not highly motivated to do so since they sell their own hosting platform.

packages/gasket-plugin-elastic-apm/CHANGELOG.md
```diff
@@ -1,18 +1,21 @@
 # `@gasket/plugin-elastic-apm`
 
-### 6.27.1
+- Add a `transactionName` lifecycle
+- Add a `transactionLabels` lifecycle
+
+## 6.27.1
```
kawikabader, line 6: @agerard-godaddy do you recall why all of our CHANGELOG.md files jump from # to ###? I feel like there was a specific reason, but I can't remember what it was.

kawikabader (approved):

packages/gasket-plugin-elastic-apm/CHANGELOG.md
agerard-godaddy: I don't recall the reason, but it has been the pattern so yea if you would, leave these at `###` for now. If we want them at `##`, then let's do a holistic change across all package changelogs so that we can remain consistent.

packages/gasket-plugin-elastic-apm/README.md
```diff
@@ -4,13 +4,13 @@ Adds Elastic APM instrumentation to your application
 
 ## Installation
 
-#### New apps
+### New apps
```
agerard-godaddy, line 7: Leave these example sections at `####` please to keep consistent. Using level 4 heading keeps example sections from crowding the docsify sidebar menu on [gasket.dev](https://gasket.dev/#/packages/gasket-plugin-elastic-apm/README).

```suggestion
#### New apps
```

packages/gasket-plugin-elastic-apm/README.md
```diff
@@ -4,13 +4,13 @@ Adds Elastic APM instrumentation to your application
...
 
 ```
 gasket create <app-name> --plugins @gasket/plugin-elastic-apm
 ```
 
-#### Existing apps
+### Existing apps
```
agerard-godaddy, line 13:
```suggestion
#### Existing apps
```

packages/gasket-plugin-elastic-apm/README.md
```diff
@@ -99,6 +99,41 @@ According to the [Elastic APM docs], the _Elastic APM agent for Node.js is a
 singleton_. This means that you can require and configure singleton in various
 hooks of your Gasket app, such as with the [init] or [middleware] lifecycles.
 
+## Lifecycles
+
+### transactionName
```
agerard-godaddy, line 104: Are `transactionName` and `transactionLabel` a little too generic for a lifecycle names? What about prefixing with _apm_ or perhaps _elastic_, i.e. `apmTransactionName`, or `elasticTransactionName`?

packages/gasket-plugin-elastic-apm/lib/middleware.js
```diff
@@ -0,0 +1,40 @@
...
+  if (!transaction) {
+    return;
+  }
+
+  const requestDetails = { req, res };
+
+  const [name, labels] = await Promise.all([
+    gasket.execWaterfall('transactionName', transaction.name, requestDetails),
+    gasket.execWaterfall('transactionLabels', {}, requestDetails)
+  ]);
```
agerard-godaddy, line 32: Somewhat related to my naming question, but couldn't these lifecycles just be condensed to a single one, and allow users to directly manage the transaction themselves without any added abstraction?

```suggestion
gasket.execWaterfall('elasticApmTransaction', transaction, requestDetails)
```

That would leave the door open for any other transaction metadata enrichment ([docs](https://www.elastic.co/guide/en/apm/guide/current/data-model-metadata.html)) without requiring additional Gasket lifecycles and supporting code.

packages/gasket-plugin-nextjs/README.md
```diff
@@ -44,7 +44,7 @@ you would set for [custom configurations][next.config] or using Next.js plugins.
 For general Webpack configurations, it is recommend to utilize features of the
 Gasket [webpack plugin].
 
-#### Example configuration
+### Example configuration
```
agerard-godaddy, line 47: Keep example sections at level 4 heading per the [markdown contrib guide](https://github.com/godaddy/gasket/blob/main/CONTRIBUTING.md#markdown-documentation) for consistency and to avoid crowding the gasket.dev sidebar nav.

packages/gasket-plugin-nextjs/README.md
```diff
@@ -58,7 +58,7 @@ module.exports = {
 }
 ```
 
-#### Example with plugins
+### Example with plugins
```
agerard-godaddy, line 61:
```suggestion
#### Example with plugins
```

packages/gasket-plugin-nextjs/README.md
```diff
@@ -184,6 +184,43 @@ module.exports = {
 }
 ```
 
+### nextPreHandling
```
agerard-godaddy, line 187: What sort of prehandling does this lifecycle solve which could not already be solved by the middleware lifecycle? I don't see any need for it in this PR code, and the given example is not convincing. Is the purpose just to provide access to the `next` instance, and if so, what do you plan to be able to do with it?

packages/gasket-plugin-nextjs/lib/next-route.js
agerard-godaddy: Do you need to let `path.join` handle the separators to keep things safe for windows?
```suggestion
      .readFile(path.join(gasket.config.root, '.next', 'routes-manifest.json'))
```

jpage-godaddy: Node.js normalizes things. Tried this on Windows:

```
> require('path').join('C:\\Windows', '.next/routes-manifest.json')
'C:\\Windows\\.next\\routes-manifest.json'
```

packages/gasket-plugin-elastic-apm/README.md
jpage-godaddy: I was wondering if we wanted to keep the lifecycle names more generic in case we have other APM providers we want to support later.

packages/gasket-plugin-elastic-apm/lib/middleware.js
jpage-godaddy: I like that; we can just pass the transaction object. Will make the change.

packages/gasket-plugin-nextjs/README.md
jpage-godaddy: Well originally I wanted to use it to label transactions I knew were about to be handled by next.js before switching to this other approach. I can remove if you'd prefer to not introduce this.

agerard-godaddy: Ok, yea it is an oddball here. Unless we have a strong use-case for it I would prefer not to introduce another lifecycle.

packages/gasket-plugin-nextjs/lib/next-route.js
jpage-godaddy: Updated it anyway

packages/gasket-plugin-elastic-apm/README.md
jpage-godaddy: Merged together into an `apmTransaction` lifecycle.

packages/gasket-plugin-elastic-apm/lib/middleware.js
jpage-godaddy: Merged, @agerard-godaddy

packages/gasket-plugin-nextjs/README.md
jpage-godaddy: Removed.

agerard-godaddy (approved):
Awesome

PR #401, opened 2022-07-27
fix: prevent next/router from changing locale in IntlProvider

gsato-godaddy:
## Summary
When using the [Next i18n Routing](https://nextjs.org/docs/advanced-features/i18n-routing) feature, the router is can cause the locale to change.  
Since the locale change is only happening on the client side, the locales are not being updating with fallbacks and mappings configured on the server.  
This occasionally results in missing translations.  

Note: Since this change prevents next/router from changing locales/translations it sort of breaks this feature:
https://nextjs.org/docs/advanced-features/i18n-routing#transition-between-locales
## Changelog
```prevent next/router from changing locale in IntlProvider```
## Test Plan
Tested locally by editing node_modules

### DIFF
--- a/packages/gasket-react-intl/src/with-intl-provider.js
+++ b/packages/gasket-react-intl/src/with-intl-provider.js
@@ -93,8 +93,6 @@ export default function withIntlProvider() {
       }
 
       const locale = localesProps?.locale ||
-        // Support for Next.js i18n routing
-        props.router?.locale || // eslint-disable-line react/prop-types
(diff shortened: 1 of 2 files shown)


agerard-godaddy (approved):
👍 Observed this fix working while paring. Gasket determines the locale to use at first request, so trying to change the locale after the fact by Next during client-routing is not suitable.

jpina1-godaddy (approved):

PR #404, opened 2022-08-01
support fastify with swagger plugin

bbetts-godaddy:
## Summary

Add a Fastify hook to the `@gasket/plugin-swagger` package to allow for Fastify to be used in place of express. 

## Changelog

- create Fastify hook

## Test Plan

Screen Shot 2022-08-01 at 1 56 53 PM


package.json
```diff
@@ -75,5 +75,8 @@
   "workspaces": [
     "./packages/*",
     "scripts/generate-docs-index"
-  ]
+  ],
+  "dependencies": {
+    "browserslist-config-godaddy": "^3.0.1"
```
jpage-godaddy, line 80: This is an internal package, no? Probably shouldn't be a dependency of this open source package.

jpage-godaddy (approved):

packages/gasket-plugin-swagger/lib/index.js
```diff
@@ -108,6 +108,30 @@ module.exports = {
...
+      timing: {
+        before: ['@gasket/plugin-nextjs']
+      },
+      handler: async function fastify(gasket, app) {
+        const { swagger, root } = gasket.config;
+        const { ui = {}, apiDocsRoute, definitionFile } = swagger;
+
+        const swaggerSpec = await loadSwaggerSpec(root, definitionFile, gasket.logger);
+
+        app.register(require('@fastify/swagger'), {
```
agerard-godaddy, line 128: 👍 Can you also delay the `require('swagger-ui-express')` to within the `express` lifecycle handler? It doesn't need to be loaded if it's not used.

packages/gasket-plugin-swagger/package.json
```diff
@@ -39,8 +39,10 @@
   },
   "homepage": "https://github.com/godaddy/gasket/tree/main/packages/gasket-plugin-swagger",
   "dependencies": {
+    "@fastify/swagger": "^6.0.0",
     "@types/swagger-ui-express": "^4.1.3",
     "express": "^4.16.3",
+    "fastify": "^4.1.0",
```
agerard-godaddy, line 45: `fastify` seems to only be used in your tests, and thus should be in `devDependencies`. It seems odd that `express` is not a dev dependency also, could it be moved, too?

agerard-godaddy (commented):
Looking great! Are there any configuration nuances between using this with express vs fastify that need to be called out in the Readme docs? Currently, it has a link to [swagger-ui-express](https://github.com/scottie1984/swagger-ui-express) options for the `ui` property. Are these the same? is there a centralized doc that would make more sense to link to?

packages/gasket-plugin-swagger/package.json
bbetts-godaddy: yeah, I don't think I see express used anywhere in the package

agerard-godaddy (approved):

PR #408, opened 2022-08-11
Remove duplicate next link in README

Zeedinstein:
Noticed a duplicate link in the readme that was unnecessary.

### DIFF
--- a/packages/gasket-plugin-nextjs/lib/index.js
+++ b/packages/gasket-plugin-nextjs/lib/index.js
@@ -256,14 +256,6 @@ module.exports = {
           after: 'next'
         },
         {
-          name: 'next',
-          method: 'exec',
-          description: 'Update the next app instance before prepare',
-          link: 'README.md#next',
-          parent: 'fastify',
-          after: 'nextConfig'
-        },
-        {
           name: 'nextFastify',
           method: 'exec',
           description: 'Access the prepared next app and fastify instance',
--- a/README.md
+++ b/README.md
@@ -102,7 +102,6 @@ Available lifecycles
 | [middleware]                                           | Add Express style middleware for Fastify                      |
 | [middleware][4]                                        | Add Express style middleware                                  |
 | [next]                                                 | Update the next app instance before prepare                   |
-| [next][5]                                              | Update the next app instance before prepare                   |
 | [nextConfig]                                           | Setup the next config                                         |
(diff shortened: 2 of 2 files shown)


jpage-godaddy (approved):

jpage-godaddy:
Is this auto-generated code, though? I wonder how it got generated as a duplicate.

jpage-godaddy (changes requested):
Actually, this is for sure generated code. Looks like the reason it was duplicated is here:

https://github.com/godaddy/gasket/blob/main/packages/gasket-plugin-nextjs/lib/index.js#L242-L265

Zeedinstein:
Sure I'll have a look at that

Zeedinstein:
@jpage-godaddy This should now only generate the one next link. Which I don't see a problem with. Unless having 2 links was done on purpose for express and fastify? But I still don't see a reason to have 2 next links.

jpage-godaddy:
> @jpage-godaddy This should now only generate the one next link. Which I don't see a problem with. Unless having 2 links was done on purpose for express and fastify? But I still don't see a reason to have 2 next links.

Hm, yeah I didn't notice until now that they had different parents, so they're not really duplicated. When the lifecycle diagram is generated, those two `next` entries are both used. It sounds like the script that generates the markdown documentation is the real issue; that needs to do the de-duplication. Finding where that bug in the documentation generation code is would be key here.

packages/gasket-plugin-nextjs/lib/index.js
```diff
@@ -255,14 +255,6 @@ module.exports = {
           parent: 'express',
           after: 'next'
         },
-        {
-          name: 'next',
```
agerard-godaddy, line 259: Good catches and thanks for being willing to contribute @Zeedinstein. As @jpage-godaddy mentioned, we would still need this entry for the graph generation.

So yea, a couple issues... we should probably only generated one or the other depending on which, either Express or Fastify, Gasket plugins are in use by an app when using the `gasket docs` command. However, that still will not address the duplication in the generated index readme for the repo here. I think the duplication will be necessary, with the best thing to do being to clarify things in the descriptions:
- `Update the Next.js app instance before preparing for Express`
- `Update the Next.js app instance before preparing for Fastify`

agerard-godaddy:
Resolved in https://github.com/godaddy/gasket/pull/472

PR #410, opened 2022-08-11
Add ci-config and ci-config-file flag options

bbetts-godaddy:
## Summary

We want to support an automated process for creating a gasket app. To bypass the user prompts, we created two flags to providing the config options (`--ci-config` and `--ci-config-file`). 

## Changelog

- add `--ci-config` and `--ci-config-file` flags

## Test Plan

Screen Shot 2022-08-11 at 3 45 47 PM

Screen Shot 2022-08-11 at 3 47 55 PM


jpage-godaddy:
Not sure about the naming; can this be generalized to something not tied to "ci" specifically, like `--prompts-json` and `--prompts-json-file`? There could be other use cases for generating a gasket app without interactive prompting.

packages/gasket-cli/src/scaffold/actions/read-config.js
```diff
@@ -0,0 +1,89 @@
+const { pluginIdentifier, flattenPresets } = require('@gasket/resolve');
```
agerard-godaddy, line 1: I don't think this file is necessary. You've duplicated logic from global-prompts, but notice in the originals, it won't actually trigger a prompt if the setting is found in the context.

As long as you've applied the provided `--config` (or file) to the create context before executing the global prompts action, it should be able to handle it.

packages/gasket-cli/src/commands/create.js
```diff
@@ -61,7 +62,17 @@ class CreateCommand extends Command {
         await loadPreset(context);
         cliVersion(context);
         applyPresetConfig(context);
-        await globalPrompts(context);
```
agerard-godaddy, line 64: Per other comment, I think you should still fire this action, even if `--config` is provided. It'll handle all the wiring up so you don't have to duplicate logic. Just be sure to read the config before.

packages/gasket-cli/src/commands/create.js
```diff
@@ -61,7 +62,17 @@ class CreateCommand extends Command {
...
+        if (ciConfig) {
+          const parsedCiConfig = JSON.parse(ciConfig);
+          Object.assign(context, parsedCiConfig);
+          await readConfig(context);
+        } else if (ciConfigFilePath) {
+          const ciConfigFile = require(ciConfigFilePath);
+          Object.assign(context, ciConfigFile);
+          await readConfig(context);
```
agerard-godaddy, line 72: Tune up your readConfig to handle either the config json string or config file and run that action early before globalPrompts.

packages/gasket-cli/src/commands/create.js
```diff
@@ -61,7 +62,17 @@ class CreateCommand extends Command {
...
+        } else {
+          await globalPrompts(context);
+        }
```
agerard-godaddy, line 75: See my other comments about still using `globalPrompts` with provide config. Introduce a new `--no-prompts` flags to purposely disable prompting. You can check if that in the `globalPrompts` handlers.

Further, there are still lifecycle driven prompts, which will come from the `promptHooks` action below. I think the simplest approach, would be to continue to execute that lifecycle, but maybe the `prompt` function a noop when `--no-prompt` as been set.

Looking at [this line](https://github.com/godaddy/gasket/blob/pfx-109/packages/gasket-cli/src/scaffold/actions/prompt-hooks.js#L64), you could probably conditional set it to a dummy function that returns an empty object if prompts are disabled. It'd be up to the user to provide the necessary `--config` if also using `--no-prompt`.

agerard-godaddy (changes requested):

packages/gasket-cli/src/commands/create.js
agerard-godaddy: You could probably tune up `inquirer.prompt` in the globalPrompts, to be have the same way; simple noop if prompts are disabled.

packages/gasket-cli/README.md
```diff
@@ -52,9 +52,9 @@ OPTIONS
                                      multiple flags (e.g. --plugins @gasket/jest --plugins example@^1.0.0)
                                      comma-separated values: --plugins=@gasket/jest,example^1.0.0
   
-  --ci-config={}                     Configuration object for the 
+  --config={}                        Configuration object for the
```
jpage-godaddy, line 55:
```suggestion
  --config={}                        JSON object that provides the values for any interactive prompts
```

packages/gasket-cli/README.md
```diff
@@ -52,9 +52,9 @@ OPTIONS
                                      multiple flags (e.g. --plugins @gasket/jest --plugins example@^1.0.0)
                                      comma-separated values: --plugins=@gasket/jest,example^1.0.0
   
-  --ci-config={}                     Configuration object for the 
+  --config={}                        Configuration object for the 
 
-  --ci-config-file
+  --config-file
```
jpage-godaddy, line 57:
```suggestion
  --config-file                        path to a JSON file that provides the values for interactive prompts
```

packages/gasket-cli/src/scaffold/utils.js
```diff
@@ -86,9 +86,30 @@ function ensureAbsolute(filepath) {
...
+ * @param {Object} configFlags - flags to read config from
+ * @param {string} configFlags.config - JSON string of config values
+ * @param {string} configFlags.configFile - path to JSON file of config values
+ */
+function readConfig(context, { config, configFile }) {
+  if (config) {
+    const parsedConfig = JSON.parse(config);
+    Object.assign(context, parsedConfig);
+  } else if (configFile) {
+    const parsedConfigFile = require(configFile);
```
jpage-godaddy, line 103: Wondering if a user might point to a relative file in a way that's not compatible with the path logic for `require`. Perhaps should be something like this instead (I'm not sure how to get the gasket root from the create context, so may need modifications):

```suggestion
    const parsedConfigFile = require(path.resolve(context.gasket.config.root, configFile));
```

bbetts-godaddy: Yeah good point!

jpage-godaddy (approved):

packages/gasket-cli/src/scaffold/actions/global-prompts.js
```diff
@@ -11,7 +11,7 @@ const { addPluginsToContext } = require('../utils');
  */
 async function chooseAppDescription(context) {
   if (!('appDescription' in context)) {
-    const { appDescription } = await inquirer.prompt([
+    const { appDescription } = context.description || await inquirer.prompt([
```
agerard-godaddy, line 14: Hmm, are you intended to normalize `context.description` with `context.appDescription` here? If `context.description` is set to a string, as in your tests, it's going to break down trying to destructure that here as `{ appDescription }`.

I don't think we need to introduced a new `context.description`. Instead, just tune up your tests to to use the existing `context.appDescription`.

packages/gasket-cli/src/scaffold/actions/global-prompts.js
```diff
@@ -31,7 +31,7 @@ async function chooseAppDescription(context) {
  * @returns {Promise} promise
  */
 async function choosePackageManager(context) {
-  const packageManager =
+  const packageManager = context.package ||
```
agerard-godaddy, line 34: Similar to the above comment. Don' introduce a new context property, just tune your tests to utilize the existing `context.packageManager`.

packages/gasket-cli/package.json
```diff
@@ -69,6 +69,7 @@
     "@oclif/plugin-help": "^5.1.10",
     "@oclif/plugin-not-found": "^2.2.4",
     "@oclif/plugin-warn-if-update-available": "^2.0.4",
+    "browserslist-config-godaddy": "^3.0.1",
```
agerard-godaddy, line 72: Why is this needed? This appears to be an internal-only package that should not be a dependency of the open source packages. CI is failing to find it as a result:
https://github.com/godaddy/gasket/runs/7883994294?check_suite_focus=true#step:4:26

packages/gasket-cli/src/scaffold/actions/global-prompts.js
```diff
@@ -142,6 +146,9 @@ const questions = [
  * @returns {Promise} promise
  */
 async function globalPrompts(context) {
+  if (!context.prompts) {
+    inquirer.prompt = () => { return {}; };
+  }
   for (var fn of questions) {
     await fn(context);
```
agerard-godaddy, line 153: Overriding a method of a required module is kinda bad practice. What would probably be safer, and more easily testable, is to conditional pass through the prompt function to the question handlers. For example:

```suggestion
  const prompt = context.prompts ? inquirer.prompt : () => ({});
  for (var fn of questions) {
    await fn(context, prompt);
```

packages/gasket-cli/src/scaffold/actions/prompt-hooks.js
```diff
@@ -61,7 +61,7 @@ async function execPluginPrompts(context, plugins = [], presets = []) {
   //
   // @see: https://github.com/SBoudrias/Inquirer.js/#inquirercreatepromptmodule---prompt-function
   //
-  const prompt = inquirer.createPromptModule();
+  const prompt = context.prompts ? inquirer.createPromptModule() : () => {};
```
agerard-godaddy, line 64: Does this return an empty object, or do you need to wrap it with parentheses?
```suggestion
  const prompt = context.prompts ? inquirer.createPromptModule() : () => ({});
```

Add a test for this if you can.

packages/gasket-cli/README.md
```diff
@@ -51,6 +51,10 @@ OPTIONS
   --plugins=plugins                  Additional plugin(s) to install. Can be set as
                                      multiple flags (e.g. --plugins @gasket/jest --plugins example@^1.0.0)
                                      comma-separated values: --plugins=@gasket/jest,example^1.0.0
+  
+  --config={}                        JSON object that provides the values for any interactive prompts
+
+  --config-file                        path to a JSON file that provides the values for interactive prompts
```
agerard-godaddy, line 57: nitpick: line this up with the rest
```suggestion
  --config-file                      path to a JSON file that provides the values for interactive prompts
```

agerard-godaddy (changes requested):

packages/gasket-cli/test/unit/scaffold/actions/global-prompts.test.js
```diff
@@ -54,40 +55,47 @@ describe('globalPrompts', () => {
     assume(promptStub).is.called(3);
   });
 
+  it('overrides inquirer prompt function with --no-prompts', async () => {
+    promptStub.returns({});
+    mockContext.prompts = false;
+    await globalPrompts(mockContext);
+    assume(mockImports.inquirer.prompt).is.a('function');
+  });
```
agerard-godaddy, line 63: I'm not sure this test is accurate or necessary.

bbetts-godaddy: Yeah I was having a hard time figuring out how to test this [line](https://github.com/godaddy/gasket/blob/pfx-109/packages/gasket-cli/src/scaffold/actions/global-prompts.js#L153)

packages/gasket-cli/src/commands/create.js
```diff
@@ -169,7 +169,21 @@ comma-separated values: --preset-path=path1,path2`,
 Instead, prefer environment variables to configure package managers
 `,
     hidden: true
-  })
+  }),
+  'config': flags.string({
+    env: 'GASKET_PLUGINS',
+    description: `If provided, do not execute prompt lifecycle`,
```
agerard-godaddy, line 175: Match up to readme description:
```suggestion
    description: `JSON object that provides the values for any interactive prompts`,
```

packages/gasket-cli/src/commands/create.js
```diff
@@ -169,7 +169,21 @@ comma-separated values: --preset-path=path1,path2`,
...
+    exclusive: ['config-file']
+  }),
+  'config-file': flags.string({
+    description: `If provided, do not execute prompt lifecycle`,
```
agerard-godaddy, line 179: Match up to readme description:
```suggestion
    description: `Path to a JSON file that provides the values for interactive prompts`,
```

agerard-godaddy (approved):
Couple of fixes then 👍

PR #411, opened 2022-08-15
Re-added nextPreHandling lifecycle.

kawikabader:
## Summary

There is a need to execute logic before next.js handles a request (ie. middleware). This code was pulled directly from a PR that was submitted by @jpage-godaddy 👏🏻 . 

## Changelog

Add a `preNextHandling` lifecycle

## Test Plan

Added unit tests.

### DIFF
--- a/packages/gasket-plugin-nextjs/README.md
+++ b/packages/gasket-plugin-nextjs/README.md
@@ -207,6 +207,24 @@ module.exports = {
 };
 ```
 
+### nextPreHandling
+
+Enables execution of custom logic just prior to an HTTP request being handed to next.js for processing. Hooks receive the request, response, and next server (not to be confused with the next function used by express-like middleware):
+
+```js
+module.exports = {
+    hooks: {
+        async nextPreHandling(gasket, {
+            req,
+            res,
+            next
+        }) {
+            await doPreRenderingLogic(req, res);
+        }
+    }
+}
+```
+
 ## Utilities
 
 This plugin adds a middleware which attaches a `getNextRoute` function to the request object. It is intended for use in server contexts where you need to know how a request will route to a next.js page. This async function returns null if the manifest could not be parsed or if the requested URL does not match a route. If a match _is_ found, an object with these properties is returned:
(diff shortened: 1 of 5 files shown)


packages/gasket-plugin-nextjs/lib/index.js
```diff
@@ -137,7 +137,15 @@ module.exports = {
         // interact with the express router we want to add a last, catch all
         // route that will activate the `next`.
         //
-        expressApp.all('*', app.getRequestHandler());
+        const nextHandler = app.getRequestHandler();
+        expressApp.all('*', async (req, res, next) => {
```
agerard-godaddy, line 141: This will need to work for Fastify apps as well

packages/gasket-plugin-nextjs/README.md
```diff
@@ -207,6 +207,24 @@ module.exports = {
...
+
+Enables execution of custom logic just prior to an HTTP request being handed to next.js for processing. Hooks receive the request, response, and next server (not to be confused with the next function used by express-like middleware):
+
+```js
+module.exports = {
+    hooks: {
+        async nextPreHandling(gasket, {
+            req,
+            res,
+            next
```
agerard-godaddy, line 220: I think this needs a different name, (perhaps `nextServer`) since `next` is the common name for middleware callbacks typically seen alongside req and res ([express docs](http://expressjs.com/en/guide/using-middleware.html#middleware.application))

kawikabader: Hmm, I see what you're saying, however, I feel like the `preHandling` verbiage is helpful in describing the lifecycle. Hmm. I wonder if we should name this lifecycle based on when it is expected to run? Or should we name it based on what the most common usage is for it, like `preRequestHandling`?

jpage-godaddy: I think this was about naming the `next` property passed to the lifecycle `nextServer`. If we did want to rename the _lifecycle_, please make sure to include `next` in the name; `preRequestHandling` is ambiguous, but `nextPreRequestHandling` or `preNextRequestHandling` would be more descriptive.

packages/gasket-plugin-nextjs/lib/index.js
kawikabader: Just updated the fastify hook to use the nextPreHandling lifecycle as well. 👍🏻

jpage-godaddy (approved):

agerard-godaddy (approved):

PR #413, opened 2022-08-18
Next.js Middleware Support

kawikabader:
## Summary

To use the new Next.js middleware functions with a custom server, we need to pass a hostname and port when creating the Next.js app.

https://nextjs.org/docs/advanced-features/custom-server

Screen Shot 2022-08-18 at 9 08 56 AM

## Changelog

Added hostname and port to Next.js app creation.

## Test Plan

Updated unit tests.

### DIFF
--- a/packages/gasket-plugin-nextjs/lib/setup-next-app.js
+++ b/packages/gasket-plugin-nextjs/lib/setup-next-app.js
@@ -10,7 +10,7 @@ const { createConfig } = require('./config');
  */
 async function setupNextApp(gasket) {
   const { exec, command, config } = gasket;
-  const { hostname, http, https, http2 } = config;
+  const { host, http, https, http2 } = config;
   const createNextApp = require('next');
   const devServer = (command.id || command) === 'local';
 
@@ -20,7 +20,7 @@ async function setupNextApp(gasket) {
   const app = createNextApp({
     dev: devServer,
     conf: await createConfig(gasket, devServer),
-    hostname,
+    host,
     port
   });
 
--- a/packages/gasket-plugin-nextjs/lib/setup-next-app.js
+++ b/packages/gasket-plugin-nextjs/lib/setup-next-app.js
@@ -10,7 +10,7 @@ const { createConfig } = require('./config');
  */
 async function setupNextApp(gasket) {
   const { exec, command, config } = gasket;
-  const { host, http, https, http2 } = config;
+  const { hostname, http, https, http2 } = config;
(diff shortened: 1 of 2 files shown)


packages/gasket-plugin-nextjs/lib/setup-next-app.js
```diff
@@ -9,13 +9,19 @@ const { createConfig } = require('./config');
...
   const devServer = (command.id || command) === 'local';
 
+  const _http = http || https || http2;
+  const port = _http.port || _http;
+
   const app = createNextApp({
     dev: devServer,
-    conf: await createConfig(gasket, devServer)
+    conf: await createConfig(gasket, devServer),
+    hostname,
```
jpage-godaddy, line 23: I don't think `hostname` here is the same as what the next server wants; hostname in the case of Gasket is the end-user's hostname, whereas the next server uses it for the IP address it'll bind the listener to. The gasket config should use the [`*.host`](https://github.com/http-party/create-servers#usage) setting for that.

Of course if we're never going to call `.listen(...)` on this it probably doesn't matter.

mmason2-godaddy (approved):

agerard-godaddy (approved):

packages/gasket-plugin-nextjs/lib/setup-next-app.js
kawikabader: I may be slightly confused by your comment. I opened [this](https://github.com/vercel/next.js/issues/39666) issue with the Next.js team thinking that it was a bug. The solution they outlined requires that we pass the hostname and port when creating the next app. After implementing this fix and testing with the Gasket Canary App, the app functioned as expected.

jpage-godaddy: It sounds like the constructor needs `hostname` because of its validation, but I doubt they actually use it unless you call `.listen()`. If you _are_ planning on using their `listen()` call then you'll need to use the `host` listener properties (the address that's being listened on), not `hostname` which is something different.

packages/gasket-plugin-nextjs/lib/setup-next-app.js
```diff
@@ -20,7 +20,7 @@ async function setupNextApp(gasket) {
   const app = createNextApp({
     dev: devServer,
     conf: await createConfig(gasket, devServer),
-    hostname,
+    host,
```
kawikabader, line 23: @jpage-godaddy Is this what you are suggesting?

PR #415, opened 2022-08-19
Fix webpack 5 compatibility issue

jpage-godaddy:
Fix runtime error when using `@gasket/react-intl` with webpack 5:

> error - ./node_modules/@gasket/react-intl/lib/with-locale-required.js:24:0
> Module not found: Can't resolve 'path'

See [this #gasket-support](https://godaddy.slack.com/archives/CABCTNQ5P/p1660934419064829) thread for details.

### DIFF
--- a/packages/gasket-react-intl/CHANGELOG.md
+++ b/packages/gasket-react-intl/CHANGELOG.md
@@ -1,5 +1,7 @@
 # `@gasket/react-intl`
 
+- Fix compatibility issue with Webpack 5 caused by top-level import of `path` module in code that could be run browser-side.
+
 ### 6.30.0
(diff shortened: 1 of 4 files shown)


jpage-godaddy:
I'm not sure how to test this in actuality. It's not clear if it's a build-time or runtime error. I think if we defer the `require` statement at runtime it won't import, but I'm not sure whether it won't still cause a build error for Webpack. Perhaps we'll need to inject webpack config to ignore `path`? Not sure if that happens automatically or not.

kawikabader (approved):

agerard-godaddy (approved):

PR #416, opened 2022-08-22
Doc how to configure cache dir for deployments

agerard-godaddy:
## Summary

We have an open PR (https://github.com/oclif/plugin-warn-if-update-available/pull/280) to gracefully fail the CLI version check, which can cause some deployment environments to fail if they cannot write to the default cache directory. In lieu of that fix, it is also possible to configure the cache directory via environment variable, as documented in this PR.

## Changelog

**@gasket/plugin-nextjs**
- Docs for cache dir gotchas during deployments

## Test Plan

Verifiable in a local gasket app using `GASKET_CACHE_DIR`

### DIFF
--- a/packages/gasket-plugin-nextjs/docs/deployment.md
+++ b/packages/gasket-plugin-nextjs/docs/deployment.md
@@ -128,3 +128,20 @@ Follow the [Docker deployment guide] to see a sample `Dockerfile`.
 
 [config]: /packages/gasket-cli/docs/configuration.md
 [Docker deployment guide]: docker-deployment.md
+
+## Gotchas
+
+### Cache directory
+
+The Gasket CLI is built upon `@oclif` and uses some plugins that need access to
+read/write to a cache directory. Based on the [oclif docs], this is configured
+to the following defaults:
+ - macOS: `~/Library/Caches/@gasket/cli`
+ - Unix: `~/.cache/@gasket/cli`
+ - Windows: `%LOCALAPPDATA%\@gasket\cli`
+
+For some deployment environments, this may need to be adjusted from the
+defaults. To override where the cache directory is for your deployment, you can
+set the `GASKET_CACHE_DIR` env variable, such as in the `Dockerfile`.
+
+[oclif docs]: https://oclif.io/docs/config


packages/gasket-plugin-nextjs/docs/deployment.md
```diff
@@ -128,3 +128,20 @@ Follow the [Docker deployment guide] to see a sample `Dockerfile`.
...
+The Gasket CLI is built upon `@oclif` and uses some plugins that need access to
+read/write to a cache directory. Based on the [oclif docs], this is configured
+to the following defaults:
+ - macOS: `~/Library/Caches/@gasket/cli`
+ - Unix: `~/.cache/@gasket/cli`
+ - Windows: `%LOCALAPPDATA%\@gasket\cli`
+
+For some deployment environments, this may need to be adjusted from the
+defaults. To override where the cache directory is for your deployment, you can
+set the `GASKET_CACHE_DIR` env variable, such as in the `Dockerfile`.
```
jpage-godaddy, line 145: Is that just the cache directory for oclif?

I like the idea of being able to set a _base_ cache directory for all things Gasket, and then specific cache directories would fall under that, like `$GASKET_CACHE_DIR/oclif`. That way we won't have the need in the future for multiple environment variables.

agerard-godaddy: Correct, it's the cache dir for oclif, and it is generated based on the cli name (`gasket`). It is not something we control directly. However, it seems like Gasket plugins and such could utilize it as a base dir for caching also.

jpage-godaddy: OK, just want to make sure it _can_ be used as a general purpose base cache directory, that oclif wouldn't cause conflicts with other caches; otherwise maybe it could be an oclif subdirectory under that base cache directory.

jpage-godaddy (approved):

packages/gasket-plugin-nextjs/docs/deployment.md
agerard-godaddy: 👍 I'll make a ticket to investigate using it for this aspect more thoroughly and we can provide some docs for plugin authors on it.

kawikabader (approved):

PR #429, opened 2022-08-30
Add middleware configuration for express plugin

jpina1-godaddy:
## Summary

Currently all server middleware runs on every request path. This has been less than ideal  in cases that call for certain middleware to be included only on particular routes. For example, middleware designed to only be run on /api/* routes would also apply to requests with the /proxy/* path. Going forward, it would be helpful for developers to have the option to configure which routes their middleware is run on in order to avoid any side effects or bugs due to unexpected middleware + request path combinations. 

## Changelog

@gasket/plugin-express
- Add check for middleware array in the gasket config

## Test Plan

- Unit tests
- Verified in a local gasket app

### DIFF
--- a/packages/gasket-plugin-express/docs/setup.md
+++ b/packages/gasket-plugin-express/docs/setup.md
@@ -106,6 +106,24 @@ module.exports = {
   }
 }
 ```
+You can configure which paths middleware will run on by adding the middleware configuration array to your local `gasket.config.js`. The array is made up of objects with the name of the middlware (Gasket plugin name) you want to configure and an array of path patterns representing the paths to match for this middleware. Pattern matching entries in the array can come in the form of path strings, path pattern strings, and/or regular expressions. Pattern matching logic follows Express's [app.use pattern matching].
+
+```js
+  middleware: [
+    {
+      plugin:'api-middleware', // Name of the Gasket plugin
+      paths: ['/api']
+    },
+    {
+      plugin:'default-middlware',
+      paths: [/\/default/]
+    },
+    {
+      plugin: 'abc-middleware'
+      paths: ['/proxy', /\/home/]
+    }
+  ]
+```
 
 <!-- LINKS -->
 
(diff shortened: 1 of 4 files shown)


jpage-godaddy (approved):

packages/gasket-plugin-express/docs/setup.md
```diff
@@ -106,10 +106,29 @@ module.exports = {
   }
 }
 ```
+You can configure which paths middleware will run on by adding the middleware configuration array to your local `gasket.config.js`. The array is made up of objects with the name of the middlware (Gasket plugin name) you want to configure and an array of path patterns representing the paths to match for this middleware. Pattern matching entries in the array can come in the form of path strings, path pattern strings, and/or regular expressions. Pattern matching logic follows Express's [app.use pattern matching].
+
+```js
+  middleware: [
+    {
+      plugin:'api-middleware', // Name of the Gasket plugin
```
agerard-godaddy, line 114: To be communicate clearly that these are Gasket plugins names, could you follow the [plugin naming convention](https://github.com/godaddy/gasket/blob/main/packages/gasket-resolve/README.md#naming-convention) for your examples, such as:
- gasket-plugin-example
- @some/gasket-plugin-example
- @another/gasket-plugin-example

packages/gasket-plugin-express/lib/index.js
```diff
@@ -76,11 +76,31 @@ module.exports = {
...
+
+        if (middleware) {
+          if (middlewareConfig) {
+            const pluginName = plugin && plugin.name ? plugin.name : '';
+            const mwConfig = middlewareConfig.find(mw => mw.plugin === pluginName);
+            if (mwConfig) {
+              middleware.paths = mwConfig.paths;
+              if (excludedRoutesRegex) {
+                middleware.paths.push(excludedRoutesRegex);
+              }
```
agerard-godaddy, line 91: 👍 nice solve to keep this backwards compatible

agerard-godaddy (approved):

PR #430, opened 2022-08-31
Add middleware configuration check for fastify plugin 

jpina1-godaddy:
## Summary

Currently all server middleware runs on every request path. This has been less than ideal  in cases that call for certain middleware to be included only on particular routes. For example, middleware designed to only be run on /api/* routes would also apply to requests with the /proxy/* path. Going forward, it would be helpful for developers to have the option to configure which routes their middleware is run on in order to avoid any side effects or bugs due to unexpected middleware + request path combinations. 

## Changelog

@gasket/plugin-fastify
- Add check for middleware array in the gasket config

## Test Plan

- Unit tests
- Verified in a local gasket app

### DIFF
--- a/packages/gasket-plugin-fastify/README.md
+++ b/packages/gasket-plugin-fastify/README.md
@@ -56,15 +56,15 @@ You can configure which paths middleware will run on by adding the middleware co
 ```js
   middleware: [
     {
-      plugin:'api-middleware', // Name of the Gasket plugin
+      plugin:'gasket-plugin-api-middleware', // Name of the Gasket plugin
       paths: ['/api']
     },
     {
-      plugin:'default-middlware',
+      plugin:'gasket-plugin-default-middlware',
       paths: [/\/default/]
     },
     {
-      plugin: 'abc-middleware'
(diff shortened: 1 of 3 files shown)


packages/gasket-plugin-fastify/README.md
```diff
@@ -50,6 +50,26 @@ module.exports = {
 }
 ```
 
+### middleware
+
+You can configure which paths middleware will run on by adding the middleware configuration array to your local `gasket.config.js`. The array is made up of objects with the name of the middlware (Gasket plugin name) you want to configure and an array of path patterns representing the paths to match for this middleware. Pattern matching entries in the array can come in the form of path strings and/or regular expressions.
+```js
+  middleware: [
+    {
+      plugin:'api-middleware', // Name of the Gasket plugin
```
agerard-godaddy, line 59: See https://github.com/godaddy/gasket/pull/429#discussion_r960780720

agerard-godaddy (approved):

kawikabader (approved):

PR #447, opened 2022-09-21
Add support for contextual thunk for locales path

agerard-godaddy:
## Summary

Some situations require changing the target locales path part based on page request details. To support this, the various locales components and HOC allow for a thunk to be passed, which will receive a context object based on the render state which can be used to determine the locales path to use.

## Changelog

**@gasket/react-intl**
- Add support for contextual thunk for locales path

**@gasket/helper-intl**
- Add support for contextual thunk for locales path

## Test Plan

- expanded unit tests
- locally tested in canary app

### DIFF
--- a/packages/gasket-helper-intl/lib/index.d.ts
+++ b/packages/gasket-helper-intl/lib/index.d.ts
@@ -29,6 +29,17 @@ export interface LocaleManifest {
  * or a URL template with a `:locale` path param to a .json file.
  */
 export type LocalePathPart = string;
+
+/**
+ * Callback which receives a context object for resolving a LocalePathPath
+ */
+export type LocalePathThunk = (context: {}) => LocalePathPart;
+
+/**
+ * Callback which receives a context object for resolving a LocalePathPath
+ */
+export type LocalePathPartOrThunk = LocalePathPart|LocalePathThunk;
+
 /**
  * URL path to a locale .json file
  */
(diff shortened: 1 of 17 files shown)


packages/gasket-helper-intl/lib/index.d.ts
```diff
@@ -29,6 +29,17 @@ export interface LocaleManifest {
  * or a URL template with a `:locale` path param to a .json file.
  */
 export type LocalePathPart = string;
+
+/**
+ * Callback which receives a context object for resolving a LocalePathPath
+ */
+export type LocalePathThunk = (context: {}) => LocalePathPart;
```
jpage-godaddy, line 36: Do we have better details on what's in the context object?

agerard-godaddy: It's fairly varied to try and type out completely. I added some notes

packages/gasket-helper-intl/lib/index.d.ts
```diff
@@ -33,7 +35,17 @@ export type LocalePathPart = string;
...
  * Callback which receives a context object for resolving a LocalePathPath
  */
-export type LocalePathThunk = (context: {}) => LocalePathPart;
+export type LocalePathThunk = (context: {
+    // Any server render
+    req?: IncomingMessage,
+    res?: OutgoingMessage
+
+    // Possible Next.js contexts
+    // @see: https://nextjs.org/docs/api-reference/data-fetching/get-static-props#context-parameter
```
jpage-godaddy, line 44: Could the properties from those docs be expanded into this type? Does next.js maybe expose a context type that can be imported?

jpage-godaddy (approved):
It'd be nice to have better types on the context, but maybe that can be expanded upon later. Having `req` already makes this useful.

mmason2-godaddy (approved):

PR #452, opened 2022-09-23
Fix types LocalePathPartOrThunk

agerard-godaddy:
## Changelog

**@gasket/react-intl**
- Allow the new path or thunk type

## Test Plan

- Updated typescript tests

### DIFF
--- a/packages/gasket-typescript-tests/test/react-intl.spec.tsx
+++ b/packages/gasket-typescript-tests/test/react-intl.spec.tsx
@@ -11,6 +11,10 @@ describe('gasket/react-intl', function () {
         initialProps: true,
         forwardRef: true
       });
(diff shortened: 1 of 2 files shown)


rmarkins-godaddy (approved):

packages/gasket-typescript-tests/test/react-intl.spec.tsx
```diff
@@ -11,6 +11,10 @@ describe('gasket/react-intl', function () {
         initialProps: true,
         forwardRef: true
       });
+      withLocaleRequired((context) => {
+        const variant = context?.req?.headers['x-locale-varaint'];
```
jpage-godaddy, line 15:
```suggestion
        const variant = context?.req?.headers['x-locale-variant'];
```

PR #461, opened 2022-09-28
Use GoDaddy eslint config for lib files

jpina1-godaddy:
## Summary

Provide linting against new javascript features not supported in older versions of node

## Changelog

- Changed linting rules on some packages to use godaddy eslint config instead of godaddy-react

## Test Plan

- Unit tests are passing


jpage-godaddy (approved):

agerard-godaddy (approved):

rxmarbles (approved):
Looks like there are conflicts from main, otherwise LGTM

PR #491, opened 2022-11-28
feat(preload locales): Adds the ability to preload locales.

jsanthosh-godaddy:
## Summary
Adds the ability to preload locales (in the middleware lifecycle) by setting `preloadLocales: true` in gasket.config.intl.

In SERP when we switched over to using gasket/plugin-intl , we noticed that the load performance can be improved by preloading the locales.
https://github.com/gdcorp-site/serp-app/blob/main/lifecycles/middleware.js#L8-L11

## Changelog
Ability to preload locales in middleware lifecycle.

## Test Plan

npm linked the package to serp-app and tested.

image

image

image

### DIFF
--- a/packages/gasket-plugin-intl/README.md
+++ b/packages/gasket-plugin-intl/README.md
@@ -60,6 +60,7 @@ required. However, these options exist to customize an app's setup.
   - `excludes` - (string[]) List of modules to ignore
 - `nextRouting` - (boolean) Enable [Next.js Routing] when used with
   [@gasket/plugin-nextjs]. (default: true)
+- `preloadLocales` (boolean) - Preloads the locale files from the manifest file in the middlewareHook.
 
 #### Example config
 
@@ -73,7 +74,8 @@ module.exports = {
     localesMap: {
       'zh-HK': 'zh-TW',
       'zh-SG': 'zh-CN'
-    }
+    },
+    preloadLocales: true
   }
 }
 ```
--- a/packages/gasket-plugin-intl/README.md
+++ b/packages/gasket-plugin-intl/README.md
@@ -60,7 +60,8 @@ required. However, these options exist to customize an app's setup.
   - `excludes` - (string[]) List of modules to ignore
 - `nextRouting` - (boolean) Enable [Next.js Routing] when used with
   [@gasket/plugin-nextjs]. (default: true)
-- `preloadLocales` (boolean) - Preloads the locale files from the manifest file in the middlewareHook.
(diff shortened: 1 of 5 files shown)


packages/gasket-plugin-intl/lib/middleware.js
```diff
@@ -29,11 +29,22 @@ function formatLocale(language) {
...
+    manifestFilename,
+    locales,
+    preloadLocales
+  } = getIntlConfig(gasket);
 
   const manifest = require(path.join(localesDir, manifestFilename));
   const localesParentDir = path.dirname(localesDir);
   const localeUtils = new LocaleUtils({ manifest, basePath });
+  if (preloadLocales) {
+    Object.keys(manifest.paths).map((localePath) => require(path.join(localesDir.slice(0, -8), localePath)));
```
agerard-godaddy, line 46: Using slice like this is risky, as `localesDir` can be configured to something else. Instead, you'd want to use `path.dirname` to get the parent directory, which `localesParentDir` has already been assigned.
```suggestion
    Object.keys(manifest.paths).map((localePath) => require(path.join(localesParentDir, localePath)));
```

packages/gasket-plugin-intl/README.md
```diff
@@ -60,6 +60,7 @@ required. However, these options exist to customize an app's setup.
   - `excludes` - (string[]) List of modules to ignore
 - `nextRouting` - (boolean) Enable [Next.js Routing] when used with
   [@gasket/plugin-nextjs]. (default: true)
+- `preloadLocales` (boolean) - Preloads the locale files from the manifest file in the middlewareHook.
```
agerard-godaddy, line 63: Nitpicking the description here to keep it more general incase we decide to move the preloading particulars to another lifecycle or something in the future.

```suggestion
- `preloadLocales` (boolean) - Preloads the locale files from the manifest at startup,
  allowing a faster first response.
```

agerard-godaddy (changes requested):
A couple small tweaks then 👍

packages/gasket-plugin-intl/README.md
jsanthosh-godaddy: done 👍

packages/gasket-plugin-intl/lib/middleware.js
jsanthosh-godaddy: Ohh good point 👍. 
Using `localesParentDir` now, Thanks @agerard-godaddy

agerard-godaddy (approved):

kawikabader (approved):

PR #492, opened 2022-12-01
[PFX-388] - Lazy Load Package Utility

mmason2-godaddy:
## Summary
- Lazy Load Package Utility
- tryResolve utility
- Associated tests

## Changelog
- add lazyLoadPackage & tryResolve utils

## Test Plan
Alls test have been updated and passed successfully

### DIFF
--- a/packages/gasket-utils/lib/lazy-load-package.js
+++ b/packages/gasket-utils/lib/lazy-load-package.js
@@ -1,6 +1,7 @@
 const fs = require('fs/promises');
 const path = require('path');
 const PackageManager = require('./package-manager');
+const tryResolve = require('./try-resolve');
 
 /**
  * Determine package manager
@@ -32,15 +33,14 @@ module.exports = async function lazyLoadPackage(dependency, gasket) {
   const { root } = gasket.config;
   const cmd = await getPkgManager(root, logger);
   const package = dependency.split('/')[0];
+  const modulePath = tryResolve(dependency, [root, __dirname]);
 
-  try {
-    require(require.resolve(`${dependency}`, { paths: [root, __dirname] }));
+  if (modulePath) {
     logger.info(`LazyLoadPackage - Package "${dependency}" already installed`);
     return require(dependency);
-  } catch (err) {
-    logger.info(`LazyLoadPackage - Package "${dependency}" not found`);
   }
 
+  logger.info(`LazyLoadPackage - Package "${dependency}" not found`);
   logger.warning(`LazyLoadPackage - installing "${package}" - save as a devDependency to avoid this`);
   const manager = new PackageManager({ packageManager: cmd, dest: root });
   await manager.exec('install', [package]);
--- a/packages/gasket-utils/lib/lazy-load-package.js
+++ b/packages/gasket-utils/lib/lazy-load-package.js
@@ -1,0 +1,79 @@
+const fs = require('fs/promises');
+const path = require('path');
+const PackageManager = require('./package-manager');
+
+/**
+ * Determine package manager
+ * @param {string} root Gasket root
+ * @param {Logger} logger Gasket logger
+ * @returns {string} Package manager cmd
+ */
+async function getPkgManager(root, logger) {
+  const yarnLock = await fs.readFile(path.join(root, 'yarn.lock'), 'utf8').catch(err => {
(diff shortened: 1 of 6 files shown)


packages/gasket-utils/lib/lazy-load-package.js
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Lazy load package - load devDependency programmatically when needed
+ * @param {string} dependency The require'ed dep needed
+ * @param {Gasket} gasket Gasket instance
+ * @returns {object} module
+ */
+module.exports = async function lazyLoadPackage(dependency, gasket) {
+  const { logger } = gasket;
+  const { root } = gasket.config;
+  const cmd = await getPkgManager(root, logger);
```
agerard-godaddy, line 34: We only need to determine the package manager if we need to install. Move this check to after attempting to require an existing module.

packages/gasket-utils/lib/lazy-load-package.js
```diff
@@ -0,0 +1,48 @@
...
+  const pkg = dependency.split('/')[0];
+  const modulePath = tryResolve(dependency, [root, __dirname]);
+
+  if (modulePath) {
+    logger.info(`LazyLoadPackage - Package "${dependency}" already installed`);
+    return require(dependency);
+  }
+
+  logger.info(`LazyLoadPackage - Package "${dependency}" not found`);
+  logger.warning(`LazyLoadPackage - installing "${pkg}" - save as a devDependency to avoid this`);
```
agerard-godaddy, line 44: This function is overly verbose right now. Let's get down to one message.

We really only need to inform the user what is going on when it appears slow (it's installing) and to let them know how to avoid it in the future.

You can let them know the package manager being used in this single log message as well.

```suggestion
  logger.info(`LazyLoadPackage - ${cmd} installing "${pkg}" - save as a devDependency to avoid this`);
```

packages/gasket-utils/lib/try-resolve.js
```diff
@@ -0,0 +1,29 @@
...
+ *   modulePath = require(modulePath)
+ * }
+ *
+ * @param {string} modulePath - Module to import
+ * @param {string[]} paths - Paths to search for the module
+ * @returns {string} module path
+ */
+function tryResolve(modulePath, paths = []) {
+  try {
+    return require.resolve(modulePath, { paths });
```
agerard-godaddy, line 20: Just match this helper's signature up to [resolve.require](https://nodejs.org/docs/latest-v16.x/api/modules.html#requireresolverequest-options)
```suggestion
function tryResolve(modulePath, options) {
  try {
    return require.resolve(modulePath, options);
```

agerard-godaddy (changes requested):

packages/gasket-utils/lib/lazy-load-package.js
```diff
@@ -0,0 +1,48 @@
...
+ * Lazy load package - load devDependency programmatically when needed
+ * @param {string} dependency The require'ed dep needed
+ * @param {Gasket} gasket Gasket instance
+ * @returns {object} module
+ */
+module.exports = async function lazyLoadPackage(dependency, gasket) {
+  const { logger } = gasket;
+  const { root } = gasket.config;
+  const cmd = await getPkgManager(root, logger);
+  const pkg = dependency.split('/')[0];
```
agerard-godaddy, line 35: This is going to break with scoped packages, i.e. lazy loading
`@my/package/deep/export/module`
would need to install
`@my/package`

agerard-godaddy: With a regex like this:
`const rePackage = /^(@[\w.]+)?\/?([\w.]+)/;`

You could probably change this to
```suggestion
  const pkg = dependency.match(rePackage)[0];
```

mmason2-godaddy:
I tend to lean on the side of more verbose than less. I removed some of the log message but left the `LazyLoadPackage` prefix. I think it's helpful when messages come through to see where they're coming from. You can run Gasket with the debug flag and probably trace the message that same way but it seems useable to include the prefix. Let me know your thoughts.

agerard-godaddy (approved):
Small tweak then 👍

jpina1-godaddy (approved):

PR #500, opened 2023-01-04
Add multi-dependency support for requireWithInstall in @gasket/utils

jpina1-godaddy:
## Summary

Update `requireWithInstall` to support multiple dependencies.

## Changelog

@gasket/utils

- Add the option for an array of dependencies to the first parameter in `requireWithInstall`.

## Test Plan

- Added unit tests and tested functionality with a local gasket app.


mmason2-godaddy (approved):
👍

packages/gasket-utils/lib/require-with-install.js
```diff
@@ -59,4 +59,23 @@ async function requireWithInstall(dependency, gasket) {
...
+async function requireWithInstall(dependencyRequest, gasket) {
+  if (!Array.isArray(dependencyRequest)) {
+    const resolvedDependency = await requireDependency(dependencyRequest, gasket);
+    return resolvedDependency;
+  }
+  const resolvedDependencyList = [];
+  for (const dependency of dependencyRequest) {
+    const resolvedDependency = await requireDependency(dependency, gasket);
+    resolvedDependencyList.push(resolvedDependency);
+  }
```
agerard-godaddy, line 77: What we need with the multi-dependency support is for the underlying `npm` call to be executed once with the list of modules to install. i.e. we should see
```
npm install my-package @scoped/package --no-save
```
and not
```
npm install my-package --no-save
npm install @scoped/package --no-save
```

The problem being, when the second install happens, npm will remove the previously installed but not saved package from node_modules.

agerard-godaddy (changes requested):

packages/gasket-utils/test/require-with-install.test.js
```diff
@@ -152,4 +152,17 @@ describe('requireWithInstall', function () {
         .equals(true);
     });
   });
+
+  describe('list of packages', function () {
+    it('allows for the first parameter to be an array of strings', async function () {
```
agerard-godaddy, line 157: Could you include in this test (or a separate one) a check that the packageManager (using `packageManagerExecStub`) is called with both packages for the install command?

agerard-godaddy (approved):

PR #542, opened 2023-01-23
[PFX-350] Upgrade to Fastify v4

bbetts-godaddy:
## Summary
Update Fastify to v4. This new version of fastify removes `app.use` functionality, however, since we are using `@fastify/middie` already, we are able to continue to use `app.user` per the [documentation here.](https://www.fastify.io/docs/latest/Guides/Migration-Guide-V4/#removed-appuse-3506)

## Changelog

- upgrade all packages to support Fastify v4

## Test Plan
Screen Shot 2023-01-23 at 3 09 56 PM
Screen Shot 2023-01-23 at 3 10 27 PM
Screen Shot 2023-01-23 at 3 11 02 PM


packages/gasket-plugin-fastify/lib/index.js
```diff
@@ -42,6 +42,8 @@ module.exports = {
       // Enable middleware for fastify@3
       await app.register(middie);
 
+      await app.register(require('@fastify/middie'));
```
agerard-godaddy, line 45: I think you need to replace `middie` with `@fastify/middie`, correct?

See: https://www.npmjs.com/package/middie

packages/gasket-plugin-fastify/package.json
```diff
@@ -50,11 +51,11 @@
     "eslint-plugin-jest": "^27.2.1",
     "eslint-plugin-json": "^3.1.0",
     "eslint-plugin-unicorn": "^44.0.0",
-    "fastify": "^3.3.0",
+    "fastify": "^4.11.0",
     "jest": "^29.3.1"
   },
   "peerDependencies": {
-    "fastify": "^3.0.0"
+    "fastify": "^4.0.0"
```
agerard-godaddy, line 58: We should keep this backwards compatible.
```suggestion
    "fastify": "^3 || ^4"
```
Unless Fastify 4 is going to require a new major Gasket release from us?

packages/gasket-plugin-nextjs/package.json
```diff
@@ -59,6 +59,7 @@
     "eslint-plugin-unicorn": "^44.0.0",
     "jest": "^29.3.1",
     "lodash.merge": "^4.6.0",
+    "middie": "^7.1.0",
```
agerard-godaddy, line 62: Probably best to go ahead and use `@fastify/middie` here

packages/gasket-plugin-swagger/test/index.test.js
```diff
@@ -285,18 +283,10 @@ describe('Swagger Plugin', function () {
...
         }
       );
     });
-
-    it('adds new routes to swagger paths', async function () {
-      await plugin.hooks.fastify.handler(mockGasket, fastify);
-      fastify.get('/hello-world', () => {});
-      await fastify.ready();
-      expect(fastify.swagger().paths).toHaveProperty('/hello-world');
-    });
```
agerard-godaddy, line 300: Why is this deleted? Is the test no longer valid?

agerard-godaddy (changes requested):

packages/gasket-plugin-fastify/package.json
bbetts-godaddy: I'm not sure if it will require a major release, there are some updates that were made in the `@gasket/plugin-swagger` package related to Fastify 4 that wouldn't work in Fastify 3 - is this something we can go over when we zoom?

kbader-godaddy:
@bbetts-godaddy is this PR still needed?

PR #560, opened 2023-03-15
Normalize app plugin names

agerard-godaddy:
## Summary

Although an atypical scenario, if a team is prototyping with app-level plugins, and adjusting timings, these were not being ordered correctly. For example, with:

```js
// ./plugins/plugin-a.js
module.exports = {
  name: 'alpha',
  hooks: {
    init(gasket) {
      console.log('INIT A');
    }
  }
};

// ./plugins/plugin-b.js
module.exports = {
  name: 'beta',
  hooks: {
    init: {
      timing: {
        before: ['alpha']
      },
      handler: function (gasket) {
        console.log('INIT B');
      }
    }
  }
};
```

when timings are set up, they are normalized to the [long form naming](https://github.com/godaddy/gasket/blob/main/packages/gasket-resolve/README.md#naming-convention), however the app plugin names themselves were not also being adjusted, so the timings would "miss".

## Changelog

**@gasket/engine**
- Fix to use long name for named app plugins

## Test Plan

Locally tested (using above examples) and updated unit tests.

### DIFF
--- a/packages/gasket-engine/lib/engine.js
+++ b/packages/gasket-engine/lib/engine.js
@@ -37,7 +37,7 @@ class PluginEngine {
       .reduce((acc, pluginInfo) => {
         let keyName;
         if (isModulePath.test(pluginInfo.name)) {
-          keyName = pluginInfo.module.name || pluginInfo.name;
+          keyName = pluginInfo.module.name ? pluginIdentifier(pluginInfo.module.name).longName : pluginInfo.name;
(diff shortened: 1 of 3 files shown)


jpage-godaddy (approved):
Oooh, nice find.

mmason2-godaddy (approved):

kawikabader (approved):

PR #590, opened 2023-04-19
Fail bad requires in gasket.config

agerard-godaddy:
## Summary

Current behavior uses tryRequire which makes it difficult to debug if the gasket.config is actually missing, or if it was a bad require in the gasket.config itself.

When a missing module is required in the gasket.config, we should hard fail the app.

## Changelog

**@gasket/resolve**
- Hard fail bad requires in gasket.config while preserving missing gasket.config behavior

## Test Plan

- Tested in local app
- Updated unit tests


mmason2-godaddy (approved):

kbader-godaddy (approved):

jpage-godaddy (approved):
Awesome

PR #593, opened 2023-05-24
Missing `glob` dependency

jpage-godaddy:
## Summary

Fix issue where `glob` isn't a declared dependency which is causing runtime errors.

[Internal GoDaddy Slack thread](https://godaddy.slack.com/archives/CABCTNQ5P/p1684942412229259) for context.

## Changelog

- Add `glob` runtime dependency to `package.json`

### DIFF
--- a/packages/gasket-plugin-intl/package.json
+++ b/packages/gasket-plugin-intl/package.json
@@ -41,6 +41,7 @@
     "@gasket/helper-intl": "^6.38.1",
     "@hapi/accept": "^6.0.0",
     "fs-extra": "^10.0.0",
+    "glob": "^7.2.0",
     "intl": "^1.2.5",
     "loader-utils": "^3.0.0",
     "lodash.merge": "^4.6.0",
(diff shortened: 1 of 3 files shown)


rmarkins-godaddy (approved):

packages/gasket-plugin-intl/package.json
```diff
@@ -41,6 +41,7 @@
     "@gasket/helper-intl": "^6.38.1",
     "@hapi/accept": "^6.0.0",
     "fs-extra": "^10.0.0",
+    "glob": "^7.2.0",
```
agerard-godaddy, line 44: Ugh, good find. I'm surprised we hadn't encountered this before.

However, this should be at `^8.0.0` to align with versions in the other plugins in the monorepo.

jpage-godaddy: Checking if it's compatible with `promisify`...

jpage-godaddy: Yep, that version also works. Updated.

kbader-godaddy (approved):

PR #598, opened 2023-06-06
git - Use `main` as default branch

chyipin:
## Summary

Use `main` as the default branch for git matching GitHub's default main branch.  Using `main` is more inclusive culturally compared to the previous default.

### 6.40.0

- Use `main` as default branch

## Test Plan

Unit tests included.


packages/gasket-plugin-git/lib/post-create.js
```diff
@@ -18,6 +18,9 @@ module.exports = {
       // Init a git repo
       await runShellCommand('git', ['init'], { cwd });
 
+      // Use `main` branch
+      await runShellCommand('git', ['checkout', '-b', 'main'], { cwd });
```
jpage-godaddy, line 22: Could you add the `-b main` to the `git init` command instead? That way we don't start out with the original branch name plus a main branch.

chyipin: `git init -b` is only available since git version 2.28.0.
This approach supports earlier versions of git; mainly to support some containers that comes with earlier versions of git.
The neat thing here is since there were no commits yet, there were no other original branch names.
We only end up with a single `main` branch.

PR #602, opened 2023-06-14
fix: Update `withGasketData` Wrapper Type

kbader-godaddy:
## Summary

`withGasketData` has an incorrect type signature which is causing typescript errors when attempting to extend the `GasketDocument`.

## Changelog

Update `withGasketData` type signature.
## Test Plan

Tests pass.


jpina1-godaddy (approved):

jpage-godaddy:
Bah, didn't know you were also working on this, but I think my pull request is more correct, and it has some tests.

https://github.com/godaddy/gasket/pull/603

PR #617, opened 2023-09-18
Fix prompt handler type

agerard-godaddy:
## Summary

More suitable type declaration which is not dependent on deep dependencies.
Fixes `error TS2305: Module '"inquirer"' has no exported member 'Inquirer'.`

## Changelog

**@gasket/cli**
- Fix prompt handler type declaration

## Test Plan

Tested change in local app.

### DIFF
--- a/packages/gasket-cli/src/index.d.ts
+++ b/packages/gasket-cli/src/index.d.ts
@@ -1,7 +1,6 @@
 import type { GasketConfigFile, MaybeAsync } from '@gasket/engine';
 import type { PackageManager } from '@gasket/utils';
 import type { Config } from '@oclif/config';
-import { Inquirer } from 'inquirer';
 
 export interface Dependencies {
   dependencies?: Record<string, string>;
@@ -196,7 +195,7 @@ declare module '@gasket/engine' {
     prompt(
       context: CreateContext,
       utils: {
-        prompt: Inquirer,
+        prompt: (prompts: Array<Record<string, any>>) => Promise<Record<string,any>>,
         addPlugins: (plugins: Array<string>) => Promise<void>
       }
     ): MaybeAsync<CreateContext>;


mmason2-godaddy (approved):
👍

rxmarbles (approved):

packages/gasket-cli/src/index.d.ts
```diff
@@ -196,7 +195,7 @@ declare module '@gasket/engine' {
     prompt(
       context: CreateContext,
       utils: {
-        prompt: Inquirer,
+        prompt: (prompts: Array<Record<string, any>>) => Promise<Record<string,any>>,
```
jpage-godaddy, line 198: This doesn't seem like a very helpful type. Can we find a better way to import the type instead?

jpage-godaddy: Or add `@types/inquirer` as a dependency?

jpage-godaddy: Looks like `@types/inquirer` has a `PromptFunction` type that maybe is what this is?

agerard-godaddy: We set up the prompt function conditionally [here](https://github.com/SBoudrias/Inquirer.js/#inquirercreatepromptmodule---prompt-function). We are using inquirer@6 in the CLI currently, which does not export types. There is a `@types/inquirer`, but then it becomes a necessary dependency for users to add, unless we make it a direct dependency of the CLI, which doesn't feel right either.

agerard-godaddy: The `PromptFunction` type is in v9 of @types/inquirer [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/inquirer/index.d.ts#L53), but it doesn't exist in v6 version [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/cfde12269ccf42b9c1db2a04e260a4f8ed4ccf62/types/inquirer/index.d.ts). Maybe I could use `PromptModule` from v6? But, either way, this will require a hard dependency of the cli on this @types/inquirer to ensure it is synced up.

agerard-godaddy: We will investigate further when upgrading
https://github.com/godaddy/gasket/issues/618

PR #623, opened 2023-09-28
Discourage first and last timing

jpage-godaddy:
## Summary

Discourage the use of `first` and `last` for lifecycle hook timings. Our own usage of these is causing difficulties with extensibility.


agerard-godaddy (approved):
Nice

packages/gasket-cli/docs/plugins.md
```diff
@@ -77,10 +77,14 @@ type Hook = HandlerFunction | {
...
+| `after` | Array of names of plugins whose hooks for the lifecycle your own hook should run after. This will guarantee that your hook will wait for the hooks of those other plugins to complete before running. |
+| `first` | Boolean indicating that your hook will run before that of any other plugin. You should avoid this whenever possible because it will prevent the ability of other hooks to run before yours, and there may be use cases for this you haven't anticipated. Instead, use the `before` property to list _known plugins_ that have hooks known to be incompatible with running before or in parallel to yours. If multiple plugins use this flag for the same lifecycle, they will run in parallel. |
+| `last` | Boolean indicating that your hook will run after those of all the other plugins. You should avoid this whenever possible because it will prevent the ability of other hooks to run after yours, and there may be use cases for this you haven't anticipated. Instead, use the `after` property to list _known plugins_ that must have their hooks complete before yours runs. If multiple plugins use this flag for the same lifecycle, they will run in parallel. |
```
agerard-godaddy, line 87: What about adjusting these slightly to relate before/after ordering with the first/last groups?
```suggestion
| `first` | Boolean indicating that your hook will run before that of any other plugin. You should avoid this whenever possible because it will prevent the ability of other hooks to run before yours, unless they also set `first`. Instead, use the `before` property to list _known plugins_ that have hooks known to be incompatible with running before or in parallel to yours. If multiple plugins use this flag for the same lifecycle, they will run in parallel unless `before` or `after` are also set. |
| `last` | Boolean indicating that your hook will run after those of all the other plugins. You should avoid this whenever possible because it will prevent the ability of other hooks to run after yours, unless they also set `last`. Instead, use the `after` property to list _known plugins_ that must have their hooks complete before yours runs. If multiple plugins use this flag for the same lifecycle, they will run in parallel unless `before` or `after` are also set. |
```

kbader-godaddy (approved):

PR #626, opened 2023-09-29
Update elastic-apm-node version.

kbader-godaddy:
## Summary

We are seeing a bad engine console warning during `npm ci`.

Screenshot 2023-09-29 at 2 24 38 PM

## Changelog

Update elastic-apm-node version.
## Test Plan

CICD passes.

### DIFF
--- a/packages/gasket-nextjs/package.json
+++ b/packages/gasket-nextjs/package.json
@@ -43,9 +43,7 @@
   "homepage": "https://github.com/godaddy/gasket/tree/main/packages/gasket-nextjs",
   "dependencies": {
     "@gasket/data": "^6.41.1",
-    "htmlescape": "^1.1.1",
-    "prop-types": "^15.8.1",
-    "react": "^18.2.0"
+    "htmlescape": "^1.1.1"
   },
   "devDependencies": {
     "@babel/cli": "^7.12.10",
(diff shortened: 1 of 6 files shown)


packages/gasket-nextjs/package.json
```diff
@@ -43,7 +43,9 @@
   "homepage": "https://github.com/godaddy/gasket/tree/main/packages/gasket-nextjs",
   "dependencies": {
     "@gasket/data": "^6.41.1",
-    "htmlescape": "^1.1.1"
+    "htmlescape": "^1.1.1",
+    "prop-types": "^15.8.1",
+    "react": "^18.2.0"
```
rmarkins-godaddy, line 48: Why did these get moved?

jpage-godaddy, line 48: Shouldn't this be a `peerDependency` instead to help avoid duplicated react instances?

packages/gasket-nextjs/package.json
```diff
@@ -66,7 +68,6 @@
     "mock-require": "^3.0.3",
     "next": "^12.0.8",
     "nyc": "^15.1.0",
-    "react": "^18.0.0",
```
jpage-godaddy, line 69: I'd leave this in `devDependencies` and pair it with an entry in `peerDependencies`.

jpage-godaddy (approved):
Elastic APM change is a definite 👍

packages/gasket-nextjs/package.json
agerard-godaddy, line 48: `react` should continue to be a devDep to allow control of version at the app level. I would say same `prop-types`.

agerard-godaddy (changes requested):

packages/gasket-nextjs/package.json
kbader-godaddy: Should I do the same for prop-types?

packages/gasket-nextjs/package.json
rmarkins-godaddy: 100% agree those should both stay as devDep

kbader-godaddy: Great catch! This actually needed to become move to `peerDependencies`. Thanks for the review!

kbader-godaddy: Yes exactly! Just pushed the update. Thanks for the review!

kbader-godaddy: Awesome. Updates have been made. Appreciate the reviews!

agerard-godaddy (approved):

PR #638, opened 2023-12-08
fix(types): add plugins to environments

radulle:
## Summary

Small update TypeScript declarations to allow plugins in environments.

### DIFF
--- a/packages/gasket-engine/lib/engine.d.ts
+++ b/packages/gasket-engine/lib/engine.d.ts
@@ -93,16 +93,18 @@ declare module '@gasket/engine' {
       ? { [K in keyof T]?: PartialRecursive<T[K]> } | undefined
       : T | undefined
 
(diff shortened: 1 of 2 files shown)


jpage-godaddy (approved):
Looks good. Would be nice to also have a TS test to go with it.

agerard-godaddy:
@radulle could you add a test to the `gasket-typescript-tests` package, probably [here](https://github.com/godaddy/gasket/blob/main/packages/gasket-typescript-tests/test/engine.spec.ts).

radulle:
> @radulle could you add a test to the `gasket-typescript-tests` package, probably [here](https://github.com/godaddy/gasket/blob/main/packages/gasket-typescript-tests/test/engine.spec.ts).

Added a test test to confirm plugins are allowed in environments.

agerard-godaddy (approved):

PR #640, opened 2023-12-14
Bring-your-own logger

jpage-godaddy:
## Summary

Gasket's logging infrastructure comprises two main parts:

1. `@gasket/plugin-log`: Manages lifecycle timing and executes the `logTransports` hook for adding extra transports to the logger configuration.
2. `@gasket/log`: Implements logging using Diagnostics (Client-side) and Winston (Server-side).

While these components work together to initialize `gasket.logger`, not all applications utilize them. Additionally, despite Winston's prevalence, we need to be aware of the numerous logging libraries in the ecosystem when adopting Gasket.

### Updates

- Create a new `@gasket/plugin-logger` to replace `@gasket/plugin-log`
- Create `@gasket/plugin-winston` to customize the default logger
- Include the new logger plugin by default
- Add a per-request logger with updateable metadata
- Update the default redux logger to use the per-request logger
- Update presets to use the winston logger

## Test Plan

Added/updated unit tests.


agerard-godaddy (commented):
I left some initial comments/questions - Obviously this is going to have some breaking changes so we will need to incorporate into our next major release.

packages/gasket-plugin-express/lib/index.js
```diff
@@ -48,6 +48,13 @@ module.exports = {
       const { root, express: { routes } = {}, http2, middleware: middlewareConfig } = config;
       const excludedRoutesRegex = config.express && config.express.excludedRoutesRegex;
       const app = http2 ? require('http2-express-bridge')(express) : express();
+      const conditionallyUse = (...middleware) => middleware.forEach(mw => {
```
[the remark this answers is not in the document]
jpage-godaddy: Renamed; hopefully is clearer now.

packages/gasket-plugin-express/lib/index.js
```diff
@@ -66,12 +73,20 @@ module.exports = {
...
 
-      if (excludedRoutesRegex) {
-        app.use(excludedRoutesRegex, cookieParser());
-      } else {
-        app.use(cookieParser());
+      function attachLogEnhancer(req) {
+        req.logger.metadata = metadata => {
+          req.logger = req.logger.child(metadata);
+          attachLogEnhancer(req);
+        };
```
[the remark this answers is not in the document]
jpage-godaddy: The idea behind this is that logging context can be added incrementally. You may have an auth middleware that adds details about the caller's groups and a later middleware that adds, say, a product ID. This is implemented by returning a child logger each time. Each created child logger needs a `metadata` method for creating the next one.

packages/gasket-plugin-express/lib/index.js
```diff
@@ -66,12 +73,20 @@ module.exports = {
...
       }
 
+      conditionallyUse((req, res, next) => {
+        req.logger = gasket.logger;
+        attachLogEnhancer(req);
+        next();
+      });
```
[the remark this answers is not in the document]
jpage-godaddy: I'm cool with that approach

packages/gasket-plugin-logger/README.md
```diff
@@ -0,0 +1,48 @@
...
+  [level: string]: (...args: Array<any>) => void,
+  child: (metadata: Object) => Logger,
+  close?: () => Promise<void> // Optional
+}
+```
+
+The `level` keys are the log levels that your logger supports. The values are functions that accept any number of arguments and write them to the log. Your logger must support, at minimum, the following levels:
+
+- `error`
+- `warn`
```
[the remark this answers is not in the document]
jpage-godaddy: I was thinking about the use case of pino which doesn't have the ability to customize levels from what I can tell. Winston comes with log levels [that match npm](https://github.com/winstonjs/winston#logging-levels), and IMO it's been a frequent gotcha where the documentation for winston doesn't make it clear whether `warn` or `warning` is correct.

We could potentially wrap the pino logger, but then it kind of feels like we're getting in the way for developers using pino that want to actually use pino. Perhaps there can be a gasket shim that monkey patches loggers?

packages/gasket-plugin-logger/src/index.ts
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable no-console */
+/* eslint-disable spaced-comment */
+
+///<reference types="@gasket/plugin-command"/>
```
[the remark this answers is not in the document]
jpage-godaddy: I'd be interested in understanding what downsides there are to authoring in TypeScript. We frequently have to fix up `.d.ts` files because they're disconnected from the actual code. Generating the `.d.ts` from TypeScript is a better way to ensure that they're correct. Also, having TypeScript validation of usage of core plugins is better than a developer honor system.

> adds some unnecessary churn for developing core plugins

Do you have an example of some of this churn?

packages/gasket-plugin-winston/src/index.ts
```diff
@@ -0,0 +1,77 @@
...
+        if (Array.isArray(transportOrTransports)) {
+          configTransports = transportOrTransports;
+        } else {
+          configTransports = [transportOrTransports];
+        }
+      } else {
+        configTransports = [new transports.Console()];
+      }
+
+      const pluginTransports = await gasket.exec('logTransports');
```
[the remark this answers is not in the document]
jpage-godaddy: I suppose it's not completely necessary, but it does make the purpose of the lifecycle more clear for those browsing the docs.

packages/gasket-plugin-logger/src/index.ts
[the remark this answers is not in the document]
agerard-godaddy: The biggest benefit is having the code that you write being the code that is executed. WIth TS, a build step is necessary to transpile to working JS. When developing locally, sure you can set up a build watcher with plugin code linked into an app. That helps, but it does not if you're trying work with a debugger. It's natural to set breakpoints in your working code, not transpiled code. Additionally, debugging or just reading the executed code is helpful if you're trying to help someone debug a problem over a call. If it's in TypeScript, going to a function definition takes you to a `.d.ts` file which is usually not helpful.

As far as churn, we struggled with the TS configs in two different monorepos with apps and plugins. Maybe friction is a better term for it. It's hard to point you to an example without an active repro case now as we did get it sorted out. It wasn't fun, though, and I'd prefer to avoid it.

I do appreciate types. Don't get me wrong. However, TypeScript types of core plugins are only helpful if the consuming app is using them, with TypeScript otherwise getting in the way during debugging and linking in development. I think there is some middle ground of using JSDocs with TypeScript for our core plugins alluded to in [this article](https://devclass.com/2023/05/11/typescript-is-not-worth-it-for-developing-libraries-says-svelte-author-as-team-switches-to-javascript-and-jsdoc/).

packages/gasket-plugin-winston/README.md
```diff
@@ -0,0 +1,177 @@
+# @gasket/plugin-winston
+
+Adds a [winston] logger instance to your Gasket instance.
```
agerard-godaddy, line 3:
```suggestion
Set up a [winston] logger instance for the Gasket logger.
```

packages/gasket-preset-api/package.json
```diff
@@ -27,7 +27,9 @@
     "@gasket/plugin-https": "^6.45.0",
     "@gasket/plugin-lint": "^6.45.0",
     "@gasket/plugin-log": "^6.45.0",
-    "@gasket/plugin-swagger": "^6.45.0"
+    "@gasket/plugin-logger": "^6.43.0",
+    "@gasket/plugin-swagger": "^6.45.0",
+    "@gasket/plugin-winston": "^6.43.0"
```
agerard-godaddy, line 32: So, we don't need both the old `log` and the new `logger`. But given `logger` will be a default plugin of the CLI, neither of these should be in the preset. Only `winston` makes sense to add.

packages/gasket-preset-nextjs/package.json
```diff
@@ -23,9 +23,11 @@
     "@gasket/plugin-express": "^6.45.0",
     "@gasket/plugin-https": "^6.45.0",
     "@gasket/plugin-log": "^6.45.0",
+    "@gasket/plugin-logger": "^6.43.0",
```
agerard-godaddy, line 26: See previous comment - we do not need these as preset plugins.

packages/gasket-typescript-tests/test/log.spec.ts
```diff
@@ -24,9 +24,8 @@ describe('gasket/log', function () {
       logger.emerg('message');
       logger.error('message');
       logger.info('message');
-      logger.warning('message');
+      logger.warn('message');
```
agerard-godaddy, line 27: So, are we keeping `@gasket/log` around? Or is it no longer necessary and going away with the `log` plugin?

If we are keeping it (and I don't know that we should) then it would need to confirm to the minimum log levels changes.

packages/gasket-log/test/client.test.js
```diff
@@ -23,7 +23,7 @@ describe('Log', function () {
 
   it('has logger for each level', function () {
     // had to change from process.stdout "write" to get passing test
-    const write = jest.spyOn(console, 'log').mockImplementation(msg => msg);
+    const write = jest.spyOn(console, 'log').mockImplementation((msg) => msg);
```
agerard-godaddy, line 26: Are we going to keep `@gasket/log`? If not, we need a ticket to remove it if it's not going to be included in this PR.

Same for removing `@gasket/plugin-log`.

packages/gasket-plugin-logger/lib/index.js
```diff
@@ -0,0 +1,73 @@
...
+      throw new Error(`Logger is missing required level: ${level}`);
+    }
+  });
+}
+
+module.exports = {
+  name,
+  hooks: {
+    async init(gasket) {
+      const loggers = await gasket.exec('createLogger');
```
mmason2-godaddy, line 30: 👍

packages/gasket-plugin-winston/README.md
```diff
@@ -0,0 +1,177 @@
...
+    ]
+  }
+}
+```
+
+## Lifecycles
+
+### logTransports
+
+To add custom logger transports, you can also hook the `logTransports` lifecycle and return a transport or an array of transports you wish to add to the logger. Here's an example gasket config and a hook that uses that config to add a FluentD traport:
```
mmason2-godaddy, line 115:
```suggestion
To add custom logger transports, you can also hook the `logTransports` lifecycle and return a transport or an array of transports you wish to add to the logger. Here's an example gasket config and a hook that uses that config to add a FluentD transport:
```

packages/gasket-typescript-tests/test/log.spec.ts
mmason2-godaddy: Seems like we won't need `gasket/log` anymore?

mmason2-godaddy:
Closing with a ticket to explore a similar approach

mmason2-godaddy:
Yeah wrong tab meant to close a different PR 🤦‍♂️

packages/gasket-plugin-intl/lib/configure.js
```diff
@@ -7,7 +7,7 @@ const moduleDefaults = {
   excludes: ['cacache', 'yargs', 'axe-core']
 };
 
-const isDefined = o => typeof o !== 'undefined';
+const isDefined = (o) => typeof o !== 'undefined';
```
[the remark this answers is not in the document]
kbader-godaddy: Yeah I think these are some prettier defaults.

kbader-godaddy:
> I'm excited for this change. Didn't see anything glaring but I would like to confirm - we will no longer need `gasket/log` correct?

That is correct.

CODEOWNERS
```diff
@@ -6,7 +6,6 @@
 
 /packages/gasket-cli/                       @indexzero, @kinetifex
```
agerard-godaddy, line 7: The codeowners file needs an overhaul. I think you may need to leave it alone in this PR and we're clean up later

packages/gasket-plugin-morgan/README.md
```diff
@@ -53,7 +51,7 @@ Defaults:
 module.exports = {
   plugins: {
     add: [
-      '@gasket/plugin-log',
+      '@gasket/plugin-logger',
```
agerard-godaddy, line 54: With logger now built-in, I don't think we need to show it being added in the examples.

packages/gasket-plugin-winston/README.md
```diff
@@ -115,18 +108,18 @@ module.exports = {
...
-   configured against the `fluentd` endpoint in the current environment.
+### winstonTransports
 
-For these scenarios the `@gasket/log` plugin exposes a `logTransports` hook:
-
-**`gasket.config.js`**
+To add custom logger transports, you can also hook the `winstonTransports`
+lifecycle and return a transport or an array of transports you wish to add to
+the logger. Here's an example gasket config and a hook that uses that config to
+add a FluentD traport:
```
mmason2-godaddy, line 118:
```suggestion
add a FluentD transport:
```
`transport`?

kbader-godaddy: _Oh yeah, you know, the traports_

packages/gasket-plugin-morgan/README.md
kbader-godaddy: DOH, missed this one.

agerard-godaddy (approved):

jpina1-godaddy (approved):

PR #643, opened 2023-12-21
Docsify bug workaround

jpage-godaddy:
## Summary

Ensure that `strip-indent@3` is floated to the top of `node_modules` to fix a [bug](https://github.com/docsifyjs/docsify/issues/2345) where `docsify-server-renderer` can require that package (which it doesn't have as a declared dependency) and receive version 4 instead, which is an ESM module.

## Changelog

Included in PR

## Test Plan

In a project where this bug was occuring, I was able to fix the issue by manually installing `strip-indent@3.0.0` as a dependency so that `strip-indent@4` was no-longer the module that was found by `docsify-server-renderer`.


ecarlson2-godaddy (approved):
LGTM

agerard-godaddy (approved):

PR #667, opened 2024-01-23
[PFX-445] - @gasket/cli Jest refactor

mmason2-godaddy:
## Changelog
 - @gasket/cli Jest refactor

## Test Plan
- Refactored.....tests


jpina1-godaddy (approved):

agerard-godaddy (approved):
Nice job with this port 🥇 

I did notice several `.mock.calls` lookups which I think we can avoid now with Jest. There are probably others like it in the other packages, so let's just tackle that in a separate cleanup pass.

kbader-godaddy (approved):

mmason2-godaddy:
> Nice job with this port 🥇
> 
> I did notice several `.mock.calls` lookups which I think we can avoid now with Jest. There are probably others like it in the other packages, so let's just tackle that in a separate cleanup pass.

I'm not sure how you would remove `mock.calls`. You restructure the test to only use `toHaveBeenCalledWith`. For this task if `args` was present I just refactored to `mock.calls`. 👍

PR #674, opened 2024-02-07
fix: adjust path on tryRequire to handle dep check

mmason2-godaddy:
## Summary
- **Why**: The `checkDevDependencies` function wasn't behaving properly due to the path used in the `tryRequire`. A require on `@docusaurus/core` will always return `null`. Opt for a `tryRequire` on the `package.json` of that file.
- **What**: Adjust path on the `tryRequire`

## Changelog
- adjust path on tryRequire to handle dep check

## Test Plan
- Update test case with the new path

### DIFF
--- a/packages/gasket-plugin-docusaurus/lib/docs-view.js
+++ b/packages/gasket-plugin-docusaurus/lib/docs-view.js
@@ -12,7 +12,7 @@ const defaultConfig = {
 
 function checkDevDependencies() {
   const preset = tryRequire('@docusaurus/preset-classic');
-  const core = tryRequire('@docusaurus/core');
+  const core = tryRequire('@docusaurus/core/lib');
   if (!preset || !core) {
     throw new Error('Missing devDependencies. Please run `npm i -D @docusaurus/core @docusaurus/preset-classic`');
   }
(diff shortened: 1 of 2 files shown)


jpage-godaddy (approved):
Seems fine as a workaround since there's not an alternative. My main worry is that if this package changes its internal structure this could break, but if this is how Docusaurus code imports from `@docusaurus/core` I guess any changes to the structure would be a breaking change anyway. Another option may be to import `@docusaurus/core/package.json` (they could break that in the future as well, though, if they messed with the exports declarations)

mmason2-godaddy:
> Seems fine as a workaround since there's not an alternative. My main worry is that if this package changes its internal structure this could break, but if this is how Docusaurus code imports from `@docusaurus/core` I guess any changes to the structure would be a breaking change anyway. Another option may be to import `@docusaurus/core/package.json` (they could break that in the future as well, though, if they messed with the exports declarations)

mmason2-godaddy:
> Seems fine as a workaround since there's not an alternative. My main worry is that if this package changes its internal structure this could break, but if this is how Docusaurus code imports from `@docusaurus/core` I guess any changes to the structure would be a breaking change anyway. Another option may be to import `@docusaurus/core/package.json` (they could break that in the future as well, though, if they messed with the exports declarations)

I'll give this a shot real quick. That's a good alternative.

jpina1-godaddy (approved):

PR #676, opened 2024-02-09
Fix and improve intl modules

agerard-godaddy:
## Summary

We had a bug where using nested `intl.modules.localesDir` was not resolving the correct package name when copying files to the public/locales dir. This PR fixes that and adds a new feature.

To allow apps to copy over varying nested locale subdirectories in packages, we now also allow a specific list of packages with optional subdirectories to be listed.

## Changelog

**@gasket/plugin-intl**
- Fix for nested localesDir scanning
- Feature to allow listing specific packages with locale subdirs

## Test Plan

Tested in local app
Screenshot 2024-02-09 at 3 29 13 PM

Screenshot 2024-02-09 at 3 00 38 PM

### DIFF
--- a/packages/gasket-plugin-intl/README.md
+++ b/packages/gasket-plugin-intl/README.md
@@ -53,7 +53,7 @@ required. However, these options exist to customize an app's setup.
   If set to `true`, the app will use the `defaultPath` as the static endpoint
   path. This option can also be set to a string, to be used as the static
   endpoint path.
-- `modules` - (boolean|object) Enable locale files collation from node modules.
+- `modules` - (boolean|object|string[]) Enable locale files collation from node modules.
   Disabled by default, enable by setting to an object with options below, or set
   to `true` to use the default options. See [Module Locales] section.
   - `localesDir` - (string) Lookup dir for module files (default: `locales`)
@@ -378,6 +378,51 @@ Because the `modules` directory is generated with each build, you may want to
 configure your SCM to ignore committing this file, such as with a `.gitignore`
 entry.
 
+## Module Files Examples
+
+Finds all node_modules with a `./locales` subdirectory.
+
+```js
+// gasket.config.js
+
+module.exports = {
+  intl: {
+    modules: true
+  }
+}
+```
+
+Find all node_modules with a `./i18n` subdirectory, excluding `my-shared-pkg`.
+
+```js
+// gasket.config.js
+
+module.exports = {
+  intl: {
+    modules: {
+        localesDir: 'i18n',
+        excludes: ['my-shared-pkg']
+    }
+  }
+}
+```
+
+Find all packages listed and their `./locales` dir or specified subdirectory.
+
+```js
+// gasket.config.js
+
+module.exports = {
+  intl: {
+    modules: [
+      'my-shared-pkg',
+      '@site/my-shared-pkg',
+      'my-other-shared-pkg/with/custom/locales-dir'
+    ]
+  }
+}
+```
+
 ## Debugging
 
 If you are experiencing difficulties seeing with locale files not working as expected, it can be helpful to enable debug logging for your gasket server via the `DEBUG` environment variable under the namespace `gasket`:
@@ -397,7 +442,7 @@ Once enabled, look for messages under the namespace `gasket:plugin:intl` and `ga
 [locales path]:#locales-path
 [locales map]:#locales-map
 [locales manifest]:#locales-manifest
-[module locales]:#locales-manifest
+[module locales]:#module-files
 [Gasket data]:#gasket-data
 [intlLocale lifecycle]:#intllocale
 [Next.js Routing]:#nextjs-routing
--- a/packages/gasket-plugin-intl/lib/build-modules.js
+++ b/packages/gasket-plugin-intl/lib/build-modules.js
@@ -5,6 +5,8 @@ const { getIntlConfig } = require('./configure');
 
 const debug = require('debug')('gasket:plugin:intl:buildModules');
 
+const rePkgParts = /^(?<name>(?:@[^/]+\/)?[\w-]+)(?<dir>\/.+)?/;
+
 class BuildModules {
   /**
    * Instantiate a builder to gather locale files
@@ -16,6 +18,11 @@ class BuildModules {
     const intlConfig = getIntlConfig(gasket);
 
     const { modules } = intlConfig;
+
+    if (Array.isArray(modules)) {
+      this._lookupModuleDirs = modules;
+    }
+
     const { excludes, localesDir } = modules;
 
     this._logger = logger;
@@ -126,11 +133,11 @@ class BuildModules {
   /**
    * Processes directories
    *
-   * @param {string[]} buildDirs - list of dirs to process
+   * @param {SrcPkgDir[]} srcPkgDirs - list of dirs to process
    */
-  async processDirs(buildDirs) {
-    for (const srcDir of buildDirs) {
-      const pkgName = await this.getPackageName(srcDir);
+  async processDirs(srcPkgDirs) {
+    for (const srcPkgDir of srcPkgDirs) {
+      const [pkgName, srcDir] = srcPkgDir;
       const tgtDir = path.join(this._outputDir, pkgName);
 
       this._logger.log(`build:locales: Updating locale files for: ${pkgName}`);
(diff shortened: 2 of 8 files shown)


packages/gasket-plugin-intl/README.md
```diff
@@ -378,6 +378,51 @@ Because the `modules` directory is generated with each build, you may want to
...
+  intl: {
+    modules: [
+      'my-shared-pkg',
+      '@site/my-shared-pkg',
+      'my-other-shared-pkg/with/custom/locales-dir'
+    ]
+  }
+}
+```
+
```
jpage-godaddy, line 425: These examples are a great addition; thanks!

packages/gasket-plugin-intl/lib/build-modules.js
```diff
@@ -5,6 +5,8 @@ const { getIntlConfig } = require('./configure');
 
 const debug = require('debug')('gasket:plugin:intl:buildModules');
 
+const rePkgParts = /^(?<name>(?:@[^/]+\/)?[\w-]+)(?<dir>\/.+)?/;
```
jpage-godaddy, line 8: Package names may have `_` as well; is that a word character? Maybe would be safer to do this:

```suggestion
const rePkgParts = /^(?<name>(?:@[^/]+\/)?[^/]+)(?<dir>\/.+)?/;
```

packages/gasket-plugin-intl/lib/build-modules.js
```diff
@@ -169,14 +178,54 @@ class BuildModules {
...
+  /**
+   * Find modules with locale directories to process
+   *
+   * @returns {SrcPkgDir[]} source package directories
+   */
+  async gatherModuleDirs() {
+    if (this._lookupModuleDirs) {
+      const promises = this._lookupModuleDirs.map(async lookupDir => {
+        const match = lookupDir.match(rePkgParts);
+        const pkgName = match.groups.name;
```
jpage-godaddy, line 190: Couldn't `match` end up being `null` if it doesn't conform to the pattern? Maybe a user-friendly error could be thrown in that case; I think it may just do a `TypeError` of some sort currently.

packages/gasket-plugin-intl/lib/build-modules.js
```diff
@@ -126,11 +133,11 @@ class BuildModules {
...
    *
-   * @param {string[]} buildDirs - list of dirs to process
+   * @param {SrcPkgDir[]} srcPkgDirs - list of dirs to process
    */
-  async processDirs(buildDirs) {
-    for (const srcDir of buildDirs) {
-      const pkgName = await this.getPackageName(srcDir);
+  async processDirs(srcPkgDirs) {
+    for (const srcPkgDir of srcPkgDirs) {
+      const [pkgName, srcDir] = srcPkgDir;
```
kbader-godaddy, line 140:
```suggestion
    for (const [pkgName, srcDir] of srcPkgDirs) {
```
Bad idea to consolidate onto one line?

packages/gasket-plugin-intl/lib/build-modules.js
```diff
@@ -148,19 +155,21 @@ class BuildModules {
...
    *
-   * @returns {string[]} directories - list of paths
+   * @returns {SrcPkgDir[]} source package directories
    */
   async discoverDirs() {
-    const dirs = await fsUtils.getDirectories(this._nodeModulesDir);
+    const pkgDirs = await fsUtils.getPackageDirs(this._nodeModulesDir);
 
-    return dirs.reduce(async (prevPromise, dir) => {
+    return pkgDirs.reduce(async (prevPromise, pkgDir) => {
```
jpage-godaddy, line 163: This is going to create a lot of in-memory Promises.

You might get better performing making `fsUtils.getPackageDirs` an async generator function, returning an async iterable, and loop through the results with `for await (const [pkgName, dir] of fsUtils.getPackageDirs(this._nodeModulesDir)) {` since you're sequentializing these anyway.

packages/gasket-plugin-intl/lib/build-modules.js
```diff
@@ -169,14 +178,54 @@ class BuildModules {
...
+          if (stat.isDirectory()) {
+            results = [pkgName, buildDir];
+          }
+        } catch (e) {
+          // skip
+        }
+        if (!results) {
+          this._logger.warning(`build:locales: locales directory not found for: ${lookupDir}`);
+        }
+        return results;
```
kbader-godaddy, line 211:
```suggestion
        try {
          const stat = await fs.lstat(buildDir);
          if (stat.isDirectory()) {
            return [pkgName, buildDir];
          }
        } catch (e) {
          // skip
        }
        this._logger.warning(
          `build:locales: locales directory not found for: ${lookupDir}`
        );
        return null;
```
Quicker return if it is a directory?

packages/gasket-plugin-intl/README.md
```diff
@@ -53,7 +53,7 @@ required. However, these options exist to customize an app's setup.
   If set to `true`, the app will use the `defaultPath` as the static endpoint
   path. This option can also be set to a string, to be used as the static
   endpoint path.
-- `modules` - (boolean|object) Enable locale files collation from node modules.
+- `modules` - (boolean|object|string[]) Enable locale files collation from node modules.
```
jpage-godaddy, line 56: Could you also change the type of modules [in the TypeScript definition](https://github.com/godaddy/gasket/blob/main/packages/gasket-plugin-intl/lib/index.d.ts#L16) as well?

kbader-godaddy (approved):

packages/gasket-plugin-intl/README.md
rvarma-godaddy: +1.

packages/gasket-plugin-intl/README.md
agerard-godaddy: Maybe you started reviewing before [this commit](https://github.com/godaddy/gasket/pull/676/commits/e9d010005dc93f25de0d5d0c47b1868fe21cf301)? - Should be all set.

packages/gasket-plugin-intl/lib/build-modules.js
agerard-godaddy: Good catch - I updated the code and tests, include various name patterns addressing your [other comment](https://github.com/godaddy/gasket/pull/676#discussion_r1484864977).

packages/gasket-plugin-intl/lib/build-modules.js
agerard-godaddy: This is how it was before, but I can take a stab at it.

jpage-godaddy: Ah yes, didn't notice that.

packages/gasket-plugin-intl/lib/build-modules.js
```diff
@@ -169,14 +177,57 @@ class BuildModules {
...
+    if (this._lookupModuleDirs) {
+      const promises = this._lookupModuleDirs.map(async lookupDir => {
+
+        const match = lookupDir.match(rePkgParts);
+        if (!match?.groups?.name) {
+          this._logger.warning(`build:locales: malformed module name: ${lookupDir}`);
+          return;
+        }
+
+        const pkgName = match.groups.name;
```
jpage-godaddy, line 195: This might be a bit DRYer

```suggestion
        const pkgName = match?.groups?.name;
        if (!pkgName) {
          this._logger.warning(`build:locales: malformed module name: ${lookupDir}`);
          return;
        }
```

packages/gasket-plugin-intl/lib/fs-utils.js
```diff
@@ -1,57 +1,64 @@
...
+ * @returns {AsyncGenerator<SrcPkgDir>} source package directories
+ */
+async function *getPackageDirs(parentDir, dirList = []) {
+  const files = await fs.readdir(parentDir);
+  for (const file of files) {
+    const filePath = path.join(parentDir, file);
+    if ((await fs.stat(filePath)).isDirectory()) {
+      const pkgName = await packageName(filePath);
+      if (pkgName) {
+        yield await Promise.resolve([pkgName, filePath]);
```
jpage-godaddy, line 39: Can be simplified to this:

```suggestion
        yield [pkgName, filePath];
```

packages/gasket-plugin-intl/lib/fs-utils.js
```diff
@@ -1,57 +1,64 @@
...
+      } else {
+        for await (const srcPkgDir of getPackageDirs(path.join(filePath, '/'), dirList)) {
+          yield await Promise.resolve(srcPkgDir);
         }
```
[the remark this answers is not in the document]
agerard-godaddy: Ah ha! Thank you!

packages/gasket-plugin-intl/lib/fs-utils.js
```diff
@@ -1,57 +1,62 @@
...
+ * @returns {AsyncGenerator<SrcPkgDir>} source package directories
+ */
+async function *getPackageDirs(parentDir, dirList = []) {
+  const files = await fs.readdir(parentDir);
+  for (const file of files) {
+    const filePath = path.join(parentDir, file);
+    if ((await fs.stat(filePath)).isDirectory()) {
+      const pkgName = await packageName(filePath);
+      if (pkgName) {
+        yield await [pkgName, filePath];
```
jpage-godaddy, line 39: You shouldn't need the await.

```suggestion
        yield [pkgName, filePath];
```

jpage-godaddy (approved):

PR #679, opened 2024-02-16
Use path instead of url in getNextRoute

schandra3-godaddy:
## Summary
`getNextRoute` regex did not work for `req.url`s with query params or anchor tags.

## Changelog
Use `req.path` instead of `req.url` in path matching in `getNextRoute`

## Test Plan
Added unit test

### DIFF
--- a/packages/gasket-plugin-nextjs/CHANGELOG.md
+++ b/packages/gasket-plugin-nextjs/CHANGELOG.md
@@ -1,6 +1,5 @@
 # `@gasket/plugin-nextjs`
 
-### 6.45.3
 
 - Use `req.path` instead of `req.url` in path matching in `getNextRoute` ([#679])
 
--- a/packages/gasket-plugin-nextjs/CHANGELOG.md
+++ b/packages/gasket-plugin-nextjs/CHANGELOG.md
@@ -1,5 +1,9 @@
 # `@gasket/plugin-nextjs`
 
+### 6.45.3
+
+- Use `req.path` instead of `req.url` in path matching in `getNextRoute` ([#679])
+
 ### 6.45.2
 
 - Add `peerDeps` ([#670])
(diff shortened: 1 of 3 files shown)


packages/gasket-plugin-nextjs/lib/next-route.js
```diff
@@ -30,7 +30,7 @@ async function getNextRoute(gasket, req) {
   const routes = await loadRoutes(gasket);
   if (routes) {
     for (const route of iterateRoutes(routes)) {
-      if (route.regex.test(req.url)) {
+      if (route.regex.test(req.path)) {
```
jpage-godaddy, line 33: One thing we need to check on is if `req.path` is valid for both Fastify and Express.

packages/gasket-plugin-nextjs/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # `@gasket/plugin-nextjs`
 
+### 6.45.3
```
jpage-godaddy, line 3: Could you remove this line? The gasket team will establish the release version numbers when they publish.

packages/gasket-plugin-nextjs/lib/next-route.js
schandra3-godaddy: Is it not available in Fastify through [Node](https://nodejs.org/docs/latest-v18.x/api/http.html#requestpath)?

packages/gasket-plugin-nextjs/CHANGELOG.md
schandra3-godaddy: Done.

packages/gasket-plugin-nextjs/lib/next-route.js
jpage-godaddy: Ah, I thought that was an express extension 👍

jpage-godaddy (approved):

PR #683, opened 2024-02-21
[PFX-238] Docusaurus Docs Site

mmason2-godaddy:
## Summary
A complete overhaul of the `gasket.dev` docs site using Docusaurus. The folder structure of the docs site has been reworked while leaving the `generate-docs-index` functionality in place.
- [Site Docs Plugin](https://github.com/godaddy/gasket/blob/PFX-238/scripts/generate-docs-index/plugins/config-plugin.js) - Handle the docs that are ran through the `docsSetup` lifecycle.
- [Copy Docs](https://github.com/godaddy/gasket/blob/PFX-238/scripts/generate-docs-index/utils/copy-site-docs.js) - Function invoked at the end of the `generate-docs-index` function. Copies root docs, transforms links, adds Front matter etc.
- [Utils](https://github.com/godaddy/gasket/tree/PFX-238/scripts/generate-docs-index/utils) - Broken down utility in functions used to new docs site generation and also broken down for simplicity.

## Todo
- [x] Deploy to GH pages and `gasket.dev` - [Workflow](https://github.com/godaddy/gasket/blob/PFX-238/.github/workflows/docs-deploy.yml) in place but disabled
- [x] [Gasket docs site](https://gasket.dev/) updated 🥳 

## Changelog
- Implement Docusaurus for the new docs site instance

## Test Plan
- [Deployed on fork](https://mmason2-godaddy.github.io/gasket-os/)


packages/gasket-engine/README.md
```diff
@@ -25,34 +25,34 @@ The `GasketAPI` object passed to hook functions has the following members:
 
 The Gasket configuration object, typically derived from `gasket.config.js`.
 
-### exec(event, ...args)
+### `exec(event, ...args)`
```
agerard-godaddy, line 28: Per [contributing guide](https://github.com/godaddy/gasket/blob/main/CONTRIBUTING.md#markdown-documentation)
> - Never use `inline code` in headings

I recall we were encountering hash link strangeness, either with GitHub or Docsify (or both). It does not appear to be an issue with GH at least now. But, any particular reason why this change? Maybe it's fine? But if so, we need to update the contributing guide if we want to move to this format.

packages/gasket-helper-intl/lib/index.js
```diff
@@ -127,11 +127,11 @@ function LocaleUtils(config) {
...
    * Fallback to the lang part of a locale or to defaultLocale.
    * Strategy is:
-   *  <locale>
-   *  <locale lang (if doesn't match default lang)>
-   *  <default locale (if a locale)>
-   *  <default lang>
-   *  null
+   *  `<locale>`
+   *  `<locale lang (if doesn't match default lang)>`
+   * `<default locale (if a locale)>`
```
agerard-godaddy, line 132: Need a space?
```suggestion
   *  `<default locale (if a locale)>`
```

scripts/generate-docs-index/index.js
```diff
@@ -21,7 +18,7 @@ async function main() {
   // copy over generated docs generated-docs
   const genSrc = path.join(__dirname, '.docs', 'docs', 'generated-docs');
   const genTgt = path.join(projectRoot, 'docs', 'generated-docs');
-  await copyDir(genSrc, genTgt);
+  await cp(genSrc, genTgt, { recursive: true });
```
agerard-godaddy, line 21: 👍

scripts/generate-docs-index/utils/add-front-matter.js
```diff
@@ -0,0 +1,79 @@
...
+ * addFrontMatter - Add front matter to the content
+ * @param {string} content The content of the file
+ * @param {string} filename The name of the file
+ * @returns {Object} The content with front matter & filename with possible extension change
+ */
+module.exports = function addFrontMatter(content, filename) {
+  // File specific transforms
+  if (frontMatterConfig[filename]?.transforms) {
+    content = frontMatterConfig[filename].transforms.reduce((content, transform) => {
+      return content.replace(transform.search, transform.replace);
```
agerard-godaddy, line 56: Eslint warning about name collision on the accumulator.
```suggestion
    content = frontMatterConfig[filename].transforms.reduce((acc, transform) => {
      return acc.replace(transform.search, transform.replace);
```

packages/gasket-engine/README.md
mmason2-godaddy: The reason for this, unfortunately, is MDX parsing the file and executing these lines as code. Using the tick allows for the MDX so see it as markdown instead of actual code which results in errors.

mmason2-godaddy: I feel like the contributing guide could remain? This is kind of a one off case IMO. I agree with the rule overall this is just what I had to do to resolve the errors.

agerard-godaddy (approved):

jpina1-godaddy (approved):

PR #692, opened 2024-02-27
[FOSS-633]: @gasket/plugin-elastic-apm 

bbetts-godaddy:
## Summary

The currentTransaction was always null in the middleware hook.  I found that when calling `apm.isStarted()`, it returned undefined. As a result, the currentTransaction was undefined meaning that the `apmTransaction` lifecycle was never executed. To work around this, I added the `apm` property to the Gasket object allowing the initial instance to be accessed in the middleware object and the lifecycle to be executed.

## Changelog

- Add `apm` to Gasket properties

## Test Plan

Screenshot 2024-02-27 at 10 28 37 AM

Screenshot 2024-02-27 at 10 29 24 AM

- Added some additional tests to the middleware hook to verify that the lifecycle is only executed when both the apm instance is started and the transaction is defined

### DIFF
--- a/packages/gasket-plugin-elastic-apm/lib/middleware.js
+++ b/packages/gasket-plugin-elastic-apm/lib/middleware.js
@@ -10,7 +10,6 @@
  */
 
 const { callbackify } = require('util');
-const apm = require('elastic-apm-node');
 
 /**
  * Middleware for customizing transactions
@@ -20,6 +19,13 @@ const apm = require('elastic-apm-node');
  * @param {Response}  res     The server response
  */
 async function customizeTransaction(gasket, req, res) {
+  const apm = gasket.apm;
+
+  if (!apm.isStarted()) {
+    gasket.logger.warning('Elastic APM has not been started properly.');
+    return;
+  }
+
   const transaction = apm.currentTransaction;
   if (!transaction) {
     return;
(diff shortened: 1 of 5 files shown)


packages/gasket-plugin-elastic-apm/lib/middleware.js
```diff
@@ -10,7 +10,6 @@
  */
 
 const { callbackify } = require('util');
-const apm = require('elastic-apm-node');
```
agerard-godaddy, line 13: Ah, I bet the inconsistent `require` statements were the root cause.

packages/gasket-plugin-elastic-apm/lib/middleware.js
```diff
@@ -20,6 +19,13 @@ const apm = require('elastic-apm-node');
  * @param {Response}  res     The server response
  */
 async function customizeTransaction(gasket, req, res) {
+  const apm = gasket.apm;
+
+  if (!apm.isStarted()) {
+    gasket.logger.warning('Elastic APM has not been started properly.');
```
agerard-godaddy, line 25: Is this log necessary or important? Seems like it would be very noisy for local development if it logs for each request.

agerard-godaddy (approved):

mmason2-godaddy (approved):

PR #694, opened 2024-03-05
[PFX-477]: Gasket CLI load dotenv by default

bbetts-godaddy:
## Summary

- Add autoloading of `.env` via provided code.

Ticket: [PFX-477](https://godaddy-corp.atlassian.net/browse/PFX-477)

## Changelog

- Add autoloading of .env via provided code.
- Add tests.

## Test Plan

Screenshot 2024-03-04 at 4 58 26 PM

`npm run test` passes


packages/gasket-plugin-express/generator/routes/index.js
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable no-unused-vars */
+require('dotenv').config();
```
agerard-godaddy, line 2: I really think this should happen earlier, as in the first line of the Gasket CLI [here](https://github.com/godaddy/gasket/blob/main/packages/gasket-cli/bin/boot).

agerard-godaddy (changes requested):

PR #695, opened 2024-03-06
[PFX-512]: Create Standard for JSDoc Types

kbader-godaddy:
## Summary

We don't currently have a documented method or examples for using JSDoc for Type Checking in the Gasket monorepo.

Wrote up a document outlining how we can achieve this. I updated the `gasket-plugin-analyze` and `gasket-plugin-express` plugins as examples. We will need to update the rest of the plugins.

## Changelog

Added JSDoc type checking, fixed types, and documented a standard.

## Test Plan

All existing tests pass.


packages/gasket-plugin-express/lib/index.d.ts
```diff
@@ -3,19 +3,32 @@ import type { Application, ErrorRequestHandler, Handler } from 'express';
...
+      plugin: string;
+      paths?: (string | RegExp)[];
+    }[];
   }
 
   export interface HookExecTypes {
-    middleware(app: Application): MaybeAsync<MaybeMultiple<Handler>>,
-    express(app: Application): MaybeAsync<void>,
-    errorMiddleware(app: Application): MaybeAsync<MaybeMultiple<ErrorRequestHandler>>
+    middleware(app: Application): MaybeAsync<MaybeMultiple<Handler> & { paths?: (string | RegExp)[] }>;
```
jpage-godaddy, line 24: Is this accurate? I'm not seeing anything else in the docs that say that `middleware` can return anything but a middleware function or an array of middleware functions. Is this saying you can have a `paths` property attached to either of those?

jpage-godaddy: Is it maybe supposed to be this instead?

```suggestion
    middleware(app: Application): MaybeAsync<MaybeMultiple<Handler & { paths?: (string | RegExp)[] }>>;
```

jpage-godaddy: Looks like maybe `paths` can be applied to "layers" which I guess is an array of handlers, so this looks ok after all. It'd be nice if these were documented.

jpage-godaddy (approved):

packages/gasket-plugin-express/lib/index.d.ts
kbader-godaddy: Yes I just stumbled onto this code yesterday and had a similar train of thought. I will add some docs.

mmason2-godaddy (approved):

bbetts-godaddy (approved):

PR #703, opened 2024-03-13
Update Gasket CLI

mmason2-godaddy:
## Summary
This PR came to light during the investigation into ESM support for Gasket. Oclif was out of date and the changes required to update and move to ESM were not feasible([See PDR](https://github.com/gdcorp-uxp/pdrs/blob/PFX-318/gasket/cli-updates.md)).

#### Extensions

Add extensions to files in the `./bin` (support future ESM work) as well as add extensions in some of the `package.json` exports. This wasn't a complete pass on extensions that will be required for ESM but it was a start and having mix should not cause issue.

#### Rename Directory

We've renamed the `src` directory to `lib` to better adhere to the convention of other plugins and the monorepo as a whole.

#### lib/bin/run.js

- File: [bin/run.js](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/bin/run.js)
- Refactor `run.js` file in the `./bin` directory
- New entry point for the cli
- Implementation using [Commander.js](https://github.com/tj/commander.js)
- [Create the](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/bin/run.js#L10) `gasket` cli and add a top-level flag for `gasket-config` path
- [Add the create command](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/bin/run.js#L22-L23)
- The `create` command [no longer makes it](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/bin/run.js#L25) into the `init` function(formerly an oclif hook)
- The `init` function is invoked if the `cmd` is not found
- The "config", formerly the `oclifConfig`, [is now defined and passed](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/bin/run.js#L28-L34) to the `init` function

#### lib/commands/create.js

- File: [commands/create.js](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/commands/create.js)
- [New and simplified signature](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/index.d.ts#L13-L21) for defining `gasket` commands
- [Create command definition](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/commands/create.js#L33)
- [Create command action definition](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/commands/create.js#L164) (command action handlers are passed the arguments and options from the definition as well as the command instance - [reference](https://github.com/tj/commander.js?tab=readme-ov-file#action-handler))

#### lib/bin/init.js

- File: [bin/init.js](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/bin/init.js)
- Replaces the Oclif `init` hook functionality and achieves the same functionality
- Function will execute when a defined command is not found
- [We need to parse](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/bin/init.js#L33) the `env` option/flag before Commander parses
- [Handle the definition of gasket env variables](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/bin/init.js#L36) based on the `env` option/flag and or variables defined in the cli
- New `commandOptions` [lifecycle](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/bin/init.js#L48-L54) - this allows for plugins to add `gasket` level options to the cli
- [Execute](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/bin/init.js#L57-L62) the renamed `getCommands` to `commands` from `init`, formerly this function executed `initOclif` which has since been removed
- [Define](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/bin/init.js#L66) `gasket.command` after the plugin commands are retrieved and execute `init` and `configure` lifecycles
- The `bin` itself, meaning the `gasket` cli instance, is passed on the config and we execute `parseAsync` [which parse arguments, options and invoke the action handler of the command](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/bin/init.js#L72)

#### lib/utils

Certain existing utilities were removed and several were added for new functionality and need with these changes.
- [env-util.js](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/utils/env-util.js)
  - `parseEnvOption` - Manually parse the `--env or --env=` option to make that value available when the `gasket.config` is loaded. This seems to be the isolated usecase where we need access to options before we parse the Commander instance.
  - `handleEnvVars` - Abstract out the setting of `gasket` environment variables for simplification and to make sure they're not overwritten.
- [process-command.js](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/utils/process-command.js)
  - `isValidCommand` - Validation of the command signature
  - `processCommand` - Process the Commander creation of a plugin-defined command
- [process-args.js](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/utils/process-args.js)
  - `isValidArg` - Validation of the argument signature
  - `processArgs` - Process the arugments to be passed into a Commander instance
- [process-options.js](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/utils/process-options.js)
  - `isValidOption` - Validation of the option signature
  - `processOptions` - Process the options to be passed into the `createOption` helper and then ultimately into the Commander instance
- [create-option.js](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/utils/create-option.js)
  - This file was created to simplify adding options manually. To make use of the complete feature set, we need to use the `addOption` method instead of the `option` method. This is required for things like `conflicts`, `hidden` and `required`. This helper function is also used in two places - [process-command](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/utils/process-command.js#L44) and [init](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/bin/init.js#L53)

#### Types

- [Config](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/index.d.ts#L5-L15) - Replaces the `oclifConfg` and is passed into `init`
- [GasketCommandDefinition](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/index.d.ts#L17-L38) - Command definition signature
- [GasketCommand](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/index.d.ts#L40-L44) - Definition after processing and the signature to be passed to Commander
- [GasketArgDefinition](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/index.d.ts#L47-L64) - Argument definition signature
- [GasketCommandArg](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/index.d.ts#L66) - Processed argument signature to be passed to Commander
- [GasketOptionDefinition](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/index.d.ts#L68C18-L95) - Option definition signature
- [GasketCommandOption](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-cli/lib/index.d.ts#L97-L104) - Processed option signature to be passed to `createOption`

#### Remove plugin-command

With the change outlined above we no longer need the `initOclif` lifecyle and the `getCommands` lifecycle is now executed from the cli. With this change we no longer have a need for `@gasket/plugin-command` and it has since been removed and `defaultPlugins` has been updated.

#### Update Plugin Commands

- [gasket-plugin-start](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-plugin-start/lib/get-commands.js#L9-L71)
- [gasket-plugin-analyze](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-plugin-analyze/lib/get-commands.js#L7-L16)
- [gasket-plugin-docs](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-plugin-docs/lib/get-commands.js#L11-L48)

#### Metrics Global Option

[Implementation](https://github.com/godaddy/gasket/blob/update-cli-v7/packages/gasket-plugin-metrics/lib/index.js#L28-L36) of the `getCommandOptions` lifecycle and the use case that inspired this change.

## Changelog
- [x] Create changelog
- Remove oclif, add Commander.js, update gasket cli

## Test Plan
- [x] Numerous tests were created and refactored.
- [x] All tests and CI/CD pass
- [x] Create tests for the lifecycles ran in the init function
- [x] `gasket create <app>`
- [x] `gasket create <app> --presets @gasket/preset-nextjs`
- [x] `gasket create <app> --presets @gasket/preset-api` 

#### Canary Publish
```
Successfully published:
- @gasket/cli@7.0.0-cli.5
```

## Additional Tickets

- [Move create command to create-gasket-app](https://godaddy-corp.atlassian.net/browse/PFX-530)
- [Add CLI update message if update is available](https://godaddy-corp.atlassian.net/browse/PFX-531)
- [Create daily testing of gasket create](https://godaddy-corp.atlassian.net/browse/PFX-529)


packages/gasket-plugin-analyze/lib/get-commands.js
```diff
@@ -2,19 +2,16 @@
...
- * @param {Object} GasketCommand - Base Gasket command to extend
  * @returns {GasketCommand} command
  */
-module.exports = function getCommands(gasket, { GasketCommand }) {
-
-  class AnalyzeCommand extends GasketCommand {
-    async gasketRun() {
-      await this.gasket.exec('build');
+module.exports = function getCommands(gasket) {
+  const AnalyzeCommand = {
```
agerard-godaddy, line 8: nit: PascalCase doesn't make sense here now, but what about just inlining this? 
```suggestion
  return {
```

packages/gasket-plugin-docs/lib/get-commands.js
```diff
@@ -6,16 +6,24 @@ const generateIndex = require('./utils/generate-index');
...
  * @param {Gasket} gasket - Gasket
- * @param {GasketCommand} GasketCommand - Base Gasket command to extend
  * @returns {GasketCommand} command
  */
-module.exports = function getCommands(gasket, { GasketCommand, flags }) {
-
-  class DocsCommand extends GasketCommand {
-    async gasketRun() {
+module.exports = function getCommands(gasket) {
+  const DocsCommand = {
```
agerard-godaddy, line 12:
```suggestion
  return {
```

packages/gasket-cli/bin/init.js
```diff
@@ -0,0 +1,82 @@
...
+    let configFile = await loadGasketConfigFile(root, env, id, options.gasketConfig);
+    if (configFile) {
+      configFile = addDefaultPlugins(configFile);
+
+      const gasket = new PluginEngine(configFile, { resolveFrom: root });
+      assignPresetConfig(gasket);
+      config.gasket = gasket;
+
+      // Add global options to the bin
+      const globalOptions = (await gasket.exec('getCommandOptions', config))
```
agerard-godaddy: 🥹🎉

packages/gasket-cli/src/index.d.ts
```diff
@@ -1,6 +1,68 @@
...
+  required?: boolean;
+  short?: string;
+  parse?: (value: string) => any;
+  type?: 'string' | 'boolean';
+  conflicts?: Array<string>; // list of option names that cannot be used together
+  hidden?: boolean; // Hide from command help output
+  default?: any; // default value
+}
+
+export interface ProccesedCLICommandOption {
```
agerard-godaddy, line 56: typo:
```suggestion
export interface ProcessedCLICommandOption {
```

But, CLI* is a bit generic. Could you rename these CLICommand* types and interfaces as GasketCommand*

packages/gasket-cli/src/index.d.ts
```diff
@@ -1,6 +1,68 @@
...
+export type ProccesedCLICommandArg = Array<string | boolean>;
+
+export interface CLICommandOption {
+  name: string;
+  description: string;
+  required?: boolean;
+  short?: string;
+  parse?: (value: string) => any;
+  type?: 'string' | 'boolean';
+  conflicts?: Array<string>; // list of option names that cannot be used together
```
agerard-godaddy, line 51: Use doc comments for property details so they show up in IDE intelliscene tips.

```suggestion
  /** list of option names that cannot be used together */
  conflicts?: Array<string>
```

packages/gasket-cli/lib/index.d.ts
```diff
@@ -111,6 +173,9 @@ export interface CreateContext {
   /** Local packages that should be linked */
   pkgLinks: Array<string>;
 
+  /** Path to npmconfig file */
+  npmconfig: string;
+
```
agerard-godaddy, line 178: This needs removed
```suggestion

```

packages/gasket-cli/lib/commands/create.js
```diff
@@ -0,0 +1,201 @@
...
+      parse: commasToArray,
+      hidden: true
+    },
+    {
+      name: 'npmconfig',
+      description: `(DEPRECATED) .npmrc to be used for npm actions in @gasket/cli'.
+      Instead, prefer environment variables to configure package managers
+      `,
+      hidden: true
+    },
```
agerard-godaddy, line 106: Should be removed completely for v7
```suggestion
```

packages/gasket-cli/lib/commands/create.js
```diff
@@ -0,0 +1,201 @@
...
+
+/**
+ * Parses comma separated option input to array
+ *
+ * @param {String} input - option argument
+ * @returns {String[]} results
+ */
+const commasToArray = input => input.split(',').map(name => name.trim());
+
+const CreateCommand = {
```
agerard-godaddy, line 33:
```suggestion
const createCommand = {
```

packages/gasket-cli/lib/bin/boot.js
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env node
-
 const { join } = require('path');
 const location = ['@gasket', 'cli', 'bin', 'run'];
```
agerard-godaddy, line 3: Should this now look in lib?
```suggestion
const location = ['@gasket', 'cli', 'lib', 'bin', 'run'];
```

packages/gasket-cli/lib/index.d.ts
```diff
@@ -1,6 +1,68 @@
 import type { GasketConfigFile, MaybeAsync } from '@gasket/engine';
 import type { PackageManager } from '@gasket/utils';
-import type { Config } from '@oclif/config';
+import type { Command } from 'commander';
+
+// ------ START OF CLI TYPES ------
```
agerard-godaddy, line 5: Are these start / end comments look needed? They seem a bit messy

packages/gasket-cli/package.json
```diff
@@ -190,21 +174,5 @@
...
-    "bin": "gasket",
-    "plugins": [
-      "@oclif/plugin-help",
-      "@oclif/plugin-not-found",
-      "@oclif/plugin-warn-if-update-available"
-    ],
-    "hooks": {
-      "init": "./src/hooks/init"
-    },
-    "warn-if-update-available": {
```
agerard-godaddy, line 205: Can you add a follow-up ticket for a custom warn-if-update-available check? It seems to encourage devs to keep current.

packages/gasket-cli/test/unit/scaffold/create-context.test.js
```diff
@@ -189,6 +189,22 @@ describe('makeCreateContext', () => {
...
+    results = makeCreateContext(argv, { npmconfig: '~/.npmconfig', presets: ['nextjs'] });
+    expect(results.npmconfig).toContain('/.npmconfig');
+    expect(path.isAbsolute(results.npmconfig)).toBeTruthy();
+  });
+
+  it('handles if npmconfig if not set', () => {
+    results = makeCreateContext(argv, { presets: ['nextjs'] });
+    expect(results.npmconfig).toBeFalsy();
+  });
+
```
agerard-godaddy, line 207: These might have crept in during rebase to v7?

```suggestion
```

mmason2-godaddy: Good catch - fixed

mmason2-godaddy:
This is PR is nearing completion but may have additional commits. I was to give a good amount of time for some extra eyes and reviews. Let me know if you have any questions.

packages/gasket-cli/lib/bin/boot.js
```diff
@@ -1,7 +1,6 @@
 #!/usr/bin/env node
-
 const { join } = require('path');
-const location = ['@gasket', 'cli', 'bin', 'run'];
+const location = ['@gasket', 'cli', 'lib', 'bin', 'run'];
```
agerard-godaddy, line 3: Sorry to churn on this... but it would probably be best for this lookup path to remain the same for backward (and forward) compatibility. For instance, if v7 cli is globally installed, it should still try to use v6 of the local app, and visa-versa.

Perhaps the package structure should look more like this:

```
./bin/run.js
./bin/boot.js
./lib/init.js
```

packages/gasket-cli/lib/bin/init.js
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env node
```
agerard-godaddy, line 1: This shouldn't be invoked directly, correct? in which case this line is unnecessary.
```suggestion
```
The `init` function is only executed by `run.js`, so could live in the lib dir (not bin) - see previous comment.

kbader-godaddy (approved):

packages/gasket-cli/lib/commands/create.js
```diff
@@ -0,0 +1,194 @@
...
+      type: 'boolean',
+      hidden: true
+    },
+    {
+      name: 'generate',
+      description: '(INTERNAL) If provided, skip the generate phase',
+      default: false,
+      type: 'boolean',
+      hidden: true
+    },
```
agerard-godaddy, line 83: This behavior is inverted now.
```
> gasket create dummy --bootstrap
⚠ Bootstrap phase skipped.
✖ Load package.json for debug. 🐲 Here be dragons!
```

With the oclif version ([code](https://github.com/godaddy/gasket/blob/1b34afeba97be5d2793c9591f69c20300cd4ef22/packages/gasket-cli/src/commands/create.js#L137-L148)), default were `true` and using `--no-bootstrap` would disable.

```
> gasket create dummy --no-bootstrap
error: unknown option '--no-bootstrap'
```

You can use `--no-*` with commander ([docs](https://www.npmjs.com/package/commander#other-option-types-negatable-boolean-and-booleanvalue)). I think it makes sense to only have the nos for these honestly.
```suggestion
    {
      name: 'no-bootstrap',
      description: '(INTERNAL) If provided, skip the bootstrap phase',
      type: 'boolean',
      hidden: true
    },
    {
      name: 'no-generate',
      description: '(INTERNAL) If provided, skip the generate phase',
      type: 'boolean',
      hidden: true
    },
```

Which means the options logic below will need adjusting also
```
const { bootstrap, generate } = options;

    if (bootstrap) {
      await bootstrap(context);
    } else {
      ora('Bootstrap phase skipped.').warn();
      if (generate) {
        await loadPkgForDebug(context);
      }
    }

    if (generate) {
```

packages/gasket-cli/lib/commands/create.js
```diff
@@ -0,0 +1,194 @@
...
+/**
+ * createCommand action
+ * @param {string} appname Required cmd arg - name of the app to create
+ * @param {object} options cmd options
+ * @param {Command} command - the command instance
+ * @returns {Promise<void>} void
+ */
+createCommand.action = async function run(appname, options, command) {
+  const argv = [appname];
+  const { bootstrap: noBootstrap, generate: noGenerate } = options;
```
agerard-godaddy, line 159: See previous comment, but effectively this would be:
```suggestion
  const { bootstrap = true, generate = true } = options;
```

packages/gasket-cli/lib/index.d.ts
```diff
@@ -187,12 +288,14 @@ export interface CreateContext {
 
 declare module '@gasket/engine' {
   export interface HookExecTypes {
-    initOclif(args: { oclifConfig: Config }): MaybeAsync<void>;
+    getCommandOptions(config: Config): MaybeAsync<Array<GasketOptionDefinition>>;
+
+    getCommands(config: Config): MaybeAsync<Array<GasketCommandDefinition> | GasketCommandDefinition>;
```
agerard-godaddy, line 293: With the pending new `actions` feature, were a prefix verb will be included, I'm wondering if these lifecycle names should just be `commands` and `commandOptions`? Leaving off a verb is more inline with the rest of existing lifecycle names.

packages/gasket-cli/lib/scaffold/actions/load-preset.js
```diff
@@ -10,16 +10,17 @@ const loader = new Loader();
  *
  * @param {[String]} rawPresets - Presets names
  * @param {String} cwd - Root path
+ * @param {Object} npmconfig - Config
  * @param {String} from from
  * @returns {PresetInfo[]} loaded presetInfos
  * @private
  */
-async function remotePresets(rawPresets, cwd, from) {
+async function remotePresets(rawPresets, cwd, npmconfig, from) {
```
agerard-godaddy, line 18: Looks like `npmconfig` made its way back into this file during a rebase.

packages/gasket-cli/lib/scaffold/create-context.js
```diff
@@ -132,6 +133,7 @@ class CreateContext {
 module.exports = function makeCreateContext(argv = [], flags = {}) {
   const appName = argv[0] || 'templated-app';
   const {
+    npmconfig,
```
agerard-godaddy, line 136:
```suggestion
```

packages/gasket-cli/lib/scaffold/fetcher.js
```diff
@@ -78,6 +78,7 @@ module.exports = class PackageFetcher {
     this.cwd = opts.cwd || process.cwd();
     this.packageName = opts.packageName;
     this.tmp = new Fetcher(opts);
+    this.npmconfig = opts.npmconfig;
```
agerard-godaddy, line 81: As in other places, `npmconfig` should be around anymore (see https://github.com/godaddy/gasket/pull/647)

packages/gasket-cli/test/unit/commands/create.test.js
```diff
@@ -25,116 +27,172 @@ mockActionStubs.installModules.update = jest.fn();
...
     expect(mockActionStubs.linkModules.update).toHaveBeenCalled();
   });
 
-  it('skips generate actions with --no-generate', async () => {
-    const cmd = new CreateCommand(['myapp', '--no-generate', '--presets=nextjs']);
-    await cmd.run();
-
+  it('skips generate actions with --generate', async () => {
+    await cmd.parseAsync(['node', 'gasket', 'create', 'myapp', '--generate', '--presets=nextjs']);
+    expect(cmdOptions.generate).toBe(true);
```
agerard-godaddy, line 127:
```suggestion
  it('skips generate actions with --no-generate', async () => {
    await cmd.parseAsync(['node', 'gasket', 'create', 'myapp', '--no-generate', '--presets=nextjs']);
    expect(cmdOptions.generate).toBe(false);
```

packages/gasket-cli/test/unit/commands/create.test.js
```diff
@@ -25,116 +27,172 @@ mockActionStubs.installModules.update = jest.fn();
...
-    expect(mockActionStubs.postCreateHooks).toHaveBeenCalled();
   });
 
-  it('skips bootstrap actions with --no-bootstrap', async () => {
-    const cmd = new CreateCommand(['myapp', '--no-bootstrap', '--presets=nextjs']);
-    await cmd.run();
-
+  it('skips bootstrap actions with --bootstrap', async () => {
+    await cmd.parseAsync(['node', 'gasket', 'create', 'myapp', '--bootstrap', '--presets=nextjs']);
+    expect(cmdOptions.bootstrap).toBe(true);
```
agerard-godaddy, line 104: This and other tests need adjust regarding previous `--no-*` option comment

```suggestion
  it('skips bootstrap actions with --no-bootstrap', async () => {
    await cmd.parseAsync(['node', 'gasket', 'create', 'myapp', '--no-bootstrap', '--presets=nextjs']);
    expect(cmdOptions.bootstrap).toBe(false);
```

agerard-godaddy (changes requested):

packages/gasket-cli/lib/index.d.ts
agerard-godaddy: I'm open to more discussion, but it seems like a good alignment opportunity now.

mmason2-godaddy: I'm open to this change. It's an involved change that touches a lot of files but I like it. Makes sense to me.

ecarlson2-godaddy (approved):

mmason2-godaddy (commented):
Ayyyeeee @agerard-godaddy

docs/generated-docs/lifecycle-graphs.md
```diff
@@ -1,7 +1,7 @@
 ```mermaid
 graph LR;
-preboot -- execWaterfall --> appEnvConfig;
-middleware -- execWaterfall --> appRequestConfig;
+* -- exec --> commandOptions;
+* -- exec --> commands;
```
mmason2-godaddy, line 4: 👍 @agerard-godaddy

packages/gasket-cli/bin/boot.js
```diff
@@ -1,8 +1,7 @@
 #!/usr/bin/env node
-
 require('dotenv').config();
 const { join } = require('path');
-const location = ['@gasket', 'cli', 'bin', 'run'];
+const location = ['@gasket', 'cli', 'lib', 'bin', 'run'];
```
agerard-godaddy, line 4:
```suggestion
const location = ['@gasket', 'cli', 'bin', 'run'];
```

agerard-godaddy: I doubled checked this locally with the extension change. It appears to work ok as long as we `require` and run has a `.js` extension. We may need to reevaluate this should we change boot/run to ESM.

agerard-godaddy (changes requested):

PR #709, opened 2024-03-22
[PFX-531] Gasket CLI: Update warning

jpina1-godaddy:
## Summary

Create similar functionality that we had with Oclif where when running the cli and an update is available we see a warning log about the newer version.

## Changelog

`@godaddy/gasket-cli`
- Added warn-if-outdated util
- Added warn-if-outdated util to run.js

## Test Plan
- Tested locally
- Added Unit tests

### DIFF
--- a/packages/gasket-cli/lib/utils/warn-if-outdated.js
+++ b/packages/gasket-cli/lib/utils/warn-if-outdated.js
@@ -1,8 +1,7 @@
-/* eslint-disable no-sync */
 const { runShellCommand } = require('@gasket/utils');
 const semver = require('semver');
 const chalk = require('chalk');
-const fs = require('fs');
+const { readFile, writeFile } = require('fs/promises');
 const path = require('path');
 
 const cachePath = path.join(__dirname, '..', '..', '.cache'); // Place at root of gasket-cli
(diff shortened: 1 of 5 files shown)


packages/gasket-cli/lib/utils/warn-if-outdated.js
```diff
@@ -0,0 +1,75 @@
+/* eslint-disable no-sync */
+const { runShellCommand } = require('@gasket/utils');
+const semver = require('semver');
+const chalk = require('chalk');
+const fs = require('fs');
```
mmason2-godaddy, line 5:
```suggestion
const { access, readFile, writeFile } = require('fs/promises');
```
You already have an `async` function so why not 🤷‍♂️

packages/gasket-cli/lib/utils/warn-if-outdated.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+module.exports = async function warnIfOutdated(pkgName, currentVersion) {
+  const currentTime = new Date().getTime();
+  let cache = {};
+
+  try {
+    const file = await readFile(cachePath, 'utf8');
+    cache = JSON.parse(file);
+  } catch (error) {
+    // console.error('Error reading cache file:', error);
```
mmason2-godaddy, line 62: Intended?

mmason2-godaddy (approved):

packages/gasket-cli/lib/utils/warn-if-outdated.js
jpina1-godaddy: Yeah, the linter doesn't like empty statements so I added this comment if it needs to be debugged in the future. I don't think it's necessary to log the error to the user if there's no cache available. Open to an alternative approach if you have another idea in mind.

mmason2-godaddy: No that makes sense 👍

bbetts-godaddy (approved):

PR #715, opened 2024-03-29
[PFX-527] JSDoc Types Part 1

kbader-godaddy:
## Summary

This PR includes the first round of updates for adding JSDoc type checking to the Gasket OS plugins. More open additional PR's as more plugins are updated.

## Changelog

Added JSDoc type checking.

## Test Plan

All current tests as well as the `typecheck:all` script pass.


packages/gasket-plugin-https/lib/index.js
```diff
@@ -62,6 +71,9 @@ async function start(gasket) {
     serverOpts.http = getPortFallback(env);
   }
 
+  /**
+   *
```
agerard-godaddy, line 75: Missing description?

packages/gasket-plugin-https/lib/index.d.ts
```diff
@@ -7,68 +7,74 @@ import type { TerminusOptions, HealthCheckError } from '@godaddy/terminus';
 
 declare module '@gasket/engine' {
   type BaseListenerConfig = {
-    port?: number,
-    host?: string,
-    timeout?: number,
-    handler?: Function
-  }
+    port?: number;
```
agerard-godaddy, line 10: semicolons here coming from an eslint rule?

agerard-godaddy (approved):

packages/gasket-plugin-https/lib/index.d.ts
kbader-godaddy: Not from an eslint rule, but from prettier. Double checked the typescriptlang.org/docs and semi-colons are what they are using as well.

mmason2-godaddy (approved):

PR #720, opened 2024-04-03
feat: initial makeGasket with actions

agerard-godaddy:
## Summary

Add `@gasket/core` package to make available the new makeGasket approach with which allows "Gaskets" to imported to app code and functionality executed by `actions`. 

## Changelog

**@gasket/core**
- Initial release
- Add `actions` lifeycycle
- Change `configure` to be syncrounous

**@gasket/engine**
- Plugin engine only takes plugins as argument
- No config setup: Config now managed completed by core package
- No plugin resolving: Plugins must be loaded and passed to engine
- Types were changed to match new signature

**@gasket/plugin-nextjs**
- Added `getNextConfig` action
- Uses `getWebpackConfig` action from webpack plugin

**@gasket/plugin-webpack**
- Added `getWebpackConfig` action, which replaces previous `initWebpack` function

## Test Plan

Basics were tested in a Next14 app using App Router


packages/gasket-utils/lib/index.d.ts
```diff
@@ -49,15 +49,11 @@ export function tryRequire(path: string): object|null;
  * @param context - Context for applying overrides
  * @param context.env - Name of environment
  * @param [context.commandId] - Name of command
- * @param [context.root] - Project root; required if using localeFile
- * @param [context.localFile] - Optional file to load relative to gasket root
  * @returns config
  */
-export function applyConfigOverrides(config: GasketConfig, { env, commandId, root, localFile }: {
+export function applyConfigOverrides(config: GasketConfigDefinition, { env, commandId }: {
```
mmason2-godaddy, line 54: `commandId` still relevant?

packages/gasket-resolve/types/config.d.ts
```diff
@@ -1,8 +1,8 @@
-import type { GasketConfigFile } from '@gasket/engine';
+import type { GasketConfigDefinition } from '@gasket/engine';
 
-export function loadGasketConfigFile(
+export function loadGasketConfigDefinition(
   root: string,
   env: string,
   commandId: string,
```
mmason2-godaddy, line 6: `commandId` comes from where?

packages/gasket-core/package.json
```diff
@@ -0,0 +1,76 @@
+{
+  "name": "@gasket/core",
+  "version": "6.45.2",
+  "description": "Entry point to setting up Gasket instances",
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
+  "type": "module",
```
mmason2-godaddy, line 7: One small step for ESM ✊

packages/gasket-core/lib/index.js
```diff
@@ -0,0 +1,85 @@
...
+import GasketEngine from '@gasket/engine';
+import { applyConfigOverrides } from '@gasket/utils';
+
+/**
+ * Get the environment to use for the gasket instance.
+ * Defaults to `development`.
+ * @returns {string} env
+ */
+function getEnvironment(
+  // flags, commandId, warn
```
mmason2-godaddy, line 14: Commented code not sure if it's intentional

mmason2-godaddy (approved):

packages/gasket-utils/lib/index.d.ts
agerard-godaddy: Dunno yet - it might still be so leaving until we make a final decision

packages/gasket-core/lib/index.js
agerard-godaddy: I left these commented out as there might still be a scenario where commands and flags could come into play. I'll add a todo so we remember to clean them up and implement - I don't want to loose track of them for now

packages/gasket-core/lib/index.js
```diff
@@ -0,0 +1,85 @@
...
+  const actionPluginMap = {};
+
+  instance.execApplySync('actions', async (plugin, handler) => {
+    const results = handler(); // The gasket parameter is automatically applied
+    if (results) {
+      Object.keys(results).forEach(actionName => {
+        if (actionPluginMap[actionName]) {
+          // eslint-disable-next-line no-console
+          console.error(
+            `Action '${actionName}' from '${plugin.name}' was registered by '${actionPluginMap[actionName]}'`
```
jpina1-godaddy, line 55: Is there any concern of naming collisions in the future between plugins that have actions with the same name?

It looks like this line takes care of the problem with actions having the same name but I'm wondering in the future when more actions are created if we will run into this issue more often and then we have to make a name change and a publish to fix it. 

We could potentially look into aliasing action names? `gasket.actions.<plugin-short-name>.<action-name>`. It's uglier but could help with collisions. 

This might not be a problem but just wanted to bring up this thought to see what the team thinks.

agerard-godaddy: Good question. I think it'll be important that the action names be unique. Aliases _could_ work possibly in application code, but we will not be able to indicate them properly with Types for intellisense and type checks. Also, some plugins will actually execute the actions of other plugins they depend on, in which case they will not know about any aliasing.

agerard-godaddy: If anything, we may consider adding a 'strict' option or something that hard fails if duplicate action names are registered. That or make it the default behavior.

jpina1-godaddy (approved):

packages/gasket-core/README.md
```diff
@@ -0,0 +1,89 @@
...
+## Lifecycles
+
+When a new Gasket is created, there are two lifecycles executed: [configure] and
+[actions].
+
+### configure
+
+The `configure` lifecycle is the first lifecycle executed when a Gasket is
+instantiated.
+This allows any registered plugin to adjust the configuration before further
```
kbader-godaddy, line 43: Would it be helpful to add a small blurb about what qualifies as a "registered plugin"?

packages/gasket-core/lib/index.js
```diff
@@ -0,0 +1,87 @@
+/* eslint-disable no-console, no-process-env */
+// @ts-check
```
kbader-godaddy, line 2: I don't think this is necessary, as the type checking should be enabled already.

packages/gasket-core/lib/index.js
```diff
@@ -0,0 +1,87 @@
...
+    return NODE_ENV;
+  }
+
+  console.warn('No env specified, falling back to "development".');
+  return 'development';
+}
+/* eslint-enable no-console, no-process-env */
+
+
+function registerActions(instance) {
```
kbader-godaddy, line 46: // TODO: Add JSDoc types

packages/gasket-core/lib/index.js
```diff
@@ -0,0 +1,87 @@
...
+    config.env = env;
+    config.root ??= process.cwd();
+
+    super(config.plugins);
+    this.command = null;
+    this.config = this.execWaterfallSync('configure', config);
+    this.actions = registerActions(this);
+  }
+}
+
```
kbader-godaddy, line 84: Also probably want to add JSDoc types here also.

packages/gasket-core/test/index.test.js
```diff
@@ -0,0 +1,178 @@
...
+
+import { jest, expect } from '@jest/globals';
+import GasketEngine from '@gasket/engine';
+import { makeGasket } from '../lib/index.js';
+
+// eslint-disable-next-line no-unused-vars
+const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
+const errorSpy = jest.spyOn(console, 'error').mockImplementation(() => {});
+
+/** @type {import('@gasket/engine').Plugin} */
```
kbader-godaddy, line 11: The tsconfig.json is currently configured to skip typechecking anything in the test directory.

packages/gasket-core/package.json
```diff
@@ -0,0 +1,76 @@
...
+  "files": [
+    "lib"
+  ],
+  "scripts": {
+    "lint": "eslint .",
+    "lint:fix": "npm run lint -- --fix",
+    "test": "cross-env NODE_OPTIONS='--unhandled-rejections=strict --experimental-vm-modules' jest",
+    "test:watch": "npm run test -- --watch",
+    "test:coverage": "npm run test -- --coverage",
+    "posttest": "npm run lint"
```
kbader-godaddy, line 17:
```suggestion
    "posttest": "npm run lint && npm run typecheck",
    "typecheck": "tsc",
    "typecheck:watch": "tsc --watch"
```

packages/gasket-engine/lib/engine.js
```diff
@@ -1,19 +1,18 @@
...
-const { Loader, pluginIdentifier } = require('@gasket/resolve');
 
 let dynamicNamingId = 0;
-const isModulePath = /^[/.]|^[a-zA-Z]:\\/;
 
-class PluginEngine {
-  constructor(config, { resolveFrom } = {}) {
-    this.config = config || {};
-    this.loader = new Loader({ resolveFrom });
+class GasketEngine {
```
kbader-godaddy, line 5: Should we move the JSDoc types and descriptions over to the engine.d.ts file and just import the @type for each of the methods here?

packages/gasket-engine/package.json
```diff
@@ -39,7 +39,6 @@
   },
   "homepage": "https://github.com/godaddy/gasket/tree/main/packages/gasket-engine",
   "dependencies": {
-    "@gasket/resolve": "^6.45.2",
     "debug": "^4.3.4"
   },
   "devDependencies": {
```
kbader-godaddy, line 44: Do we want to add typescript and the tsc typechecking scripts here yet?

kbader-godaddy (approved):

packages/gasket-engine/lib/engine.js
agerard-godaddy: Yea, I want to do a significant refactor of types here so let's save for a following PR

packages/gasket-engine/package.json
agerard-godaddy: Yes, but not in this PR

packages/gasket-core/test/index.test.js
agerard-godaddy: Ah yes good to know - I don't think it hurts to have these here for reference at least. Perhaps we should enable type checks for some tests?

PR #721, opened 2024-04-03
[PFX-530] Gasket CLI: Create command to create-gasket-app

bbetts-godaddy:
## Summary

[PFX-530](https://godaddy-corp.atlassian.net/browse/PFX-530)

- Remove create command from `@gasket/cli` and have all new apps created with `create-gasket-app`

## Changelog

- remove create command from the CLI
- add create command to `create-gasket-app` under the hood

## Test Plan

Screenshot 2024-04-03 at 2 25 39 PM

- Able to create a new application using create-gasket-app
- All tests are passing


packages/gasket-cli/bin/run.js
```diff
@@ -22,11 +21,13 @@ const gasketBin = program
...
   const cmd = process.argv[2];
   const isFlag = cmd && cmd.charAt(0) === '-';
-  const { command, hidden, isDefault } = processCommand(CreateCommand);
   await warnIfOutdated(pkg.name, pkg.version);
-  gasketBin.addCommand(command, { hidden, isDefault });
 
-  if (cmd === 'create' || isFlag) return await gasketBin.parseAsync();
+  if (cmd === 'create') {
+    return console.warn('The create command has been removed. Use npx create-gasket-app instead.');
+  }
```
agerard-godaddy, line 28: 👍

agerard-godaddy (approved):
Nice work migrating this over!

mmason2-godaddy (approved):

packages/create-gasket-app/lib/utils/warn-if-outdated.js
```diff
@@ -0,0 +1,76 @@
+const { runShellCommand } = require('@gasket/utils');
+const semver = require('semver');
+const chalk = require('chalk');
+const { readFile, writeFile } = require('fs/promises');
+const path = require('path');
+
+const cachePath = path.join(__dirname, '..', '..', '.cache'); // Place at root of gasket-cli
```
jpina1-godaddy, line 7: I think the `.cahe` file is getting added to the root of the `create-gasket-app` directory with this change and not to the `gasket-cli` directory. Not sure if we still want the `.cache` file to be saved to the `gasket-cli` directory or we are moving it to the `create-gasket-app` directory. 

If we keep `.cache` here we need to update the `.gitignore` to ignore the new cache path.

jpina1-godaddy: Just realized that `warn-if-outdated` is duplicated in `create-gasket-app` and `gasket-cli`. 

Maybe it's worth a ticket to consolidate this file in one place and share that between the two packages? @agerard-godaddy @mmason2-godaddy

mmason2-godaddy: You shouldn't need the warning if outdated at all right? Although the naming of the CLI in the index is `gasket` this is essentially a completely separate CLI.

packages/create-gasket-app/lib/index.js
```diff
@@ -1,16 +1,29 @@
...
+const CreateCommand = require('./commands/create');
 
-const { fork } = require('child_process');
-const path = require('path');
+const { processCommand, logo, warnIfOutdated } = require('./utils');
 
-const [, , ...args] = process.argv;
+// Create Gasket CLI
+const gasketBin = program
+  .name('gasket')
```
mmason2-godaddy, line 12:
```suggestion
  .name('create-gasket-app')
```

packages/create-gasket-app/lib/utils/warn-if-outdated.js
mmason2-godaddy: @jpina1-godaddy Let's create a ticket to clean up the duplicate `/utils` folder. Not sure where is should go exactly.... maybe have `create-gasket-app` depend on `@gasket/cli` and update the exports of the CLI package.

jpina1-godaddy: Sounds good. I created a ticket in the backlog to give this issue a little more thought and to potentially remove this from `create-gasket-app` if it's not needed.

mmason2-godaddy: Awesome thank you. It may be needed but it's a weird edge case. If you every `npm i -g create-gasket-app` then you'd want it so we probably should have it given that's possible scenario. The ticket to create exported utils from the CLI is still relevant.

PR #722, opened 2024-04-04
[PFX-550]: Fix lint script failure warning after creating app with api preset

bbetts-godaddy:
## Summary

- Ticket: [PFX-550](https://godaddy-corp.atlassian.net/browse/PFX-550)

This error: `Errors encountered running script: 'lint:fix'` is displayed when creating a gasket app with the `@godaddy/gasket-preset-api` and `@gasket/preset-api` presets. The error is due to `next` not being defined in the generated file for the express routes. 

## Changelog

- Add eslint disable 

## Test Plan

- All tests pass
- The error is no longer 
Screenshot 2024-04-04 at 11 15 17 AM
occurring when creating a gasket app with the presets

### DIFF
--- a/packages/gasket-plugin-express/generator/routes/index.js
+++ b/packages/gasket-plugin-express/generator/routes/index.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-unused-vars */
 module.exports = (app) => {
   {{#if hasSwaggerPlugin}}
   /**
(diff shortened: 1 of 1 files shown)


packages/gasket-plugin-express/generator/routes/index.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
```
agerard-godaddy, line 1: Would it be better to just remove `next` from the route handler? I don't think it should be there.

bbetts-godaddy: Yeah! I'll remove that now.

mmason2-godaddy (approved):

PR #725, opened 2024-04-09
[PFX-553] - startServer action, nextjs plugin files, comment out plugin start, createLogger action, update tests

mmason2-godaddy:
This PR will probably be adjusted after the [base branch](https://github.com/godaddy/gasket/pull/720) is merged.

## Summary
- `startServer` action
- `devProxy` lifecycle & config
- `createLogger` action
- Generate new files from the nextjs plugin
- Tweak added deps from nextjs plugin
- Add prompt for server type
- Temporarily comment out plugin start functionality

## Todo
- [ ] Update docs - it seems premature to spend so much time on docs given we have so many moving parts
- [ ] Update nextjs versions to 14 - unsure if we're ready to make this change or if it will happen in this PR
- [ ] Tweak preset dependencies - this will need to happen unsure if it will happen in this PR

## Changelog
- startServer action, nextjs plugin files, comment out plugin start, createLogger action, update tests

## Test Plan
- Updated tests
- Tested locally with a ton of linking


packages/gasket-plugin-logger/lib/index.js
```diff
@@ -23,24 +23,28 @@ function verifyLoggerLevels(logger) {
...
-        throw new Error(
-          'Multiple plugins are hooking createLogger. Only one logger is supported.'
-        );
-      } else {
-        verifyLoggerLevels(loggers[0]);
-        gasket.logger = loggers[0];
-      }
+    actions(gasket) {
+      return {
+        createLogger: async function () {
```
agerard-godaddy, line 28: Not crazy about this. We shouldn't have to call this at every entry.

Maybe setting up the logger is a case for keeping around `init`, but making it synchronous.

Alternatively, we make the action return the logger for an on demand approach.

```
myHook(gasket) {
  gasket.actions.getLogger().debug('yo')
}
```

Otherwise, let's attach as before during `init`.

```
myHook(gasket) {
  gasket.logger.debug('yo')
}
```


Current lifecycle flow is

1. `configure`
2. `actions`

So, if we want the logger to be available for use in `configure`, then that suggest `init` will still be important.

1. `init`
2. `configure`
3. `actions`

Or else, we register actions before configure, to make a getLogger action available

1. `actions`
2. `configure`

agerard-godaddy: I'm leaning towards keeping `init`. As much as I'd prefer to avoid attaching a bunch of stuff to the `gasket` instance, it is less of a code change across existing plugins for now.

Regardless, I'm trying to think how/if config would affect what actions are registered, and I don't that scenario exists. Sure, config make affect how an action behaves, but they should be predicably registered otherwise that's going to be a code and TS types nightmare.

So I'm going to propose we have

1. `init`
2. `actions`
3. `configure`

Let's leave `gasket.logger` for v7 and if we feel like we want be strict with actions instead of attachment, then we deprecate `gasket.logger` and move to `gasket.actions.getLogger()` for v8.

packages/gasket-plugin-nextjs/lib/create.js
```diff
@@ -0,0 +1,147 @@
...
+ * @property {Files} files - The Gasket Files API.
+ * @property {generatorDir} - The directory of the generator.
+ * @property {testPlugin} - Selected test plugin from prompt
+ */
+function createTestFiles({ files, generatorDir, testPlugin }) {
+  const frameworks = ['jest', 'mocha', 'cypress'];
+  frameworks.forEach((tester) => {
+    if (
+      testPlugin &&
+      pluginIdentifier(testPlugin).longName === `@gasket/plugin-${tester}`
```
agerard-godaddy, line 28: We need to move away from supporting short names and using `pluginIdentifier` so we can completely axe the `@gasket/register`. We'll need a different pattern here, but I'm fine if it's handled in a different PR.

packages/gasket-plugin-nextjs/lib/setup-next-app.js
```diff
@@ -18,14 +18,13 @@ function getPortFallback(env = '') {
  * @private
  */
 async function setupNextApp(gasket) {
-  const { exec, command, config } = gasket;
+  const { exec, config } = gasket;
   const { hostname, http, https, http2, env } = config;
   const createNextApp = require('next');
-  const devServer = (command.id || command) === 'local';
+  const devServer = Boolean(process.env.GASKET_DEV); // TODO document GASKET_DEV
```
agerard-godaddy, line 24: Is `GASKET_DEV` going to be official then? I don't have a better suggestion at the moment.

mmason2-godaddy: I think so I also don't have a better alternative. It's simple and explicit.

mmason2-godaddy: Documentation should be covered in our docs tickets

packages/gasket-plugin-nextjs/lib/create.js
mmason2-godaddy: I overlooked this. With the removal of resolve this needs to change sooner rather than later. I'll update.

packages/gasket-plugin-logger/lib/index.js
agerard-godaddy: Adding init and changing the order in https://github.com/godaddy/gasket/pull/728

packages/gasket-plugin-logger/lib/index.js
```diff
@@ -43,6 +43,11 @@ module.exports = {
         gasket.logger = loggers[0];
       }
     },
+    actions(gasket) {
+      return {
+        getLogger: () => gasket.logger
```
agerard-godaddy, line 48: A good start 👍

packages/gasket-plugin-nextjs/generator/next/server.mjs
```diff
@@ -0,0 +1,7 @@
+import gasket from './gasket.mjs';
+
+async function main() {
+  await gasket.actions.startServer();
+}
```
agerard-godaddy, line 5: Is awaiting necessary here?
```suggestion
function main() {
  gasket.actions.startServer();
}
```

mmason2-godaddy: That function is async... https://github.com/godaddy/gasket/pull/725/files#diff-1ad8f0a34128ce187e4dff25274f52b81634ad2ea45db507006ebab859f21dceR76-R83

We'd need to update every lifecycle to change that. There's I think 8 or so in that file.

agerard-godaddy: No I just mean in this server.mjs file. No real reason for the await. For example if this were TS we might write it as 
```
void gasket.actions.startServer();
```

mmason2-godaddy: Ahh yes I get what you're saying. Also should these be `.js`?

agerard-godaddy (approved):

packages/gasket-plugin-logger/lib/index.js
jpina1-godaddy, line 48: Why do we need a `getLogger` action? If we have access to the gasket object when calling `gasket.actions.getLogger`, why not use `gasket.logger` straight away?

jpina1-godaddy (approved):

packages/gasket-plugin-logger/lib/index.js
mmason2-godaddy: It's just exposed as a utility for now... the actual practicality of it may be up for debate for sure.

PR #727, opened 2024-04-11
[PFX-350]: Upgrade to Fastify v4

bbetts-godaddy:
## Summary

- Ticket: [PFX-350](https://godaddy-corp.atlassian.net/browse/PFX-350)

This PR includes updates to the BYO logger to be compatible with the `pino` logger used by Fastify. 

## Changelog

- Upgrade to Fastify v4

## Test Plan

- Creating an app with the fastify plugin locally builds and runs as expected.  

### Tests Passing
Screen Shot 2023-01-23 at 3 09 56 PM
Screen Shot 2023-01-23 at 3 10 27 PM
Screen Shot 2023-01-23 at 3 11 02 PM


packages/gasket-plugin-swagger/test/index.test.js
```diff
@@ -298,18 +298,21 @@ describe('Swagger Plugin', function () {
...
+        swagger: { data: true }
+      });
     });
 
     it('adds new routes to swagger paths', async function () {
       await plugin.hooks.fastify.handler(mockGasket, fastify);
       fastify.get('/hello-world', () => { });
       await fastify.ready();
-      expect(fastify.swagger().paths).toHaveProperty('/hello-world');
+      // expect(fastify.swagger().paths).toHaveProperty('/hello-world');
```
jpina1-godaddy, line 315: Does this need to be uncommented?

jpina1-godaddy (approved):

packages/gasket-plugin-swagger/test/index.test.js
bbetts-godaddy: Removed this test because it no longer works with `@fastify/swagger-ui`. I tested the swagger changes in a new api preset and it works there

packages/gasket-plugin-logger/lib/index.js
```diff
@@ -8,12 +8,14 @@ function createChildLogger(parent, metadata) {
     error: (...args) => console.error(...args, metadata),
     info: (...args) => console.info(...args, metadata),
     warn: (...args) => console.warn(...args, metadata),
+    fatal: (...args) => console.error(...args, metadata),
+    trace: (...args) => console.trace(...args, metadata),
```
agerard-godaddy, line 12: I believe we landed on only providing `console` log method/levels in the core logger plugin. That way we don't have to chase other implementation possibilities.

How are these other levels used by fastify? If the extra methods are required, could you perhaps create a child logger for it?

```
const logger = gasket.logger.child({
  fatal: gasket.logger.fatal ?? gasket.logger.error,
  trace: gasket.logger.trace ?? gasket.logger.debug
})

somethingThatNeedsMoreLevels(logger)
```

agerard-godaddy (changes requested):

mmason2-godaddy (approved):

packages/gasket-plugin-fastify/lib/utils.js
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * A helper function to ensure that all the required log levels
+ * for a Fastify server are present.
+ * @type {import('./index').alignLogger}
+ */
+function alignLogger(logger) {
+  const fastifyLogger = logger;
+  ['fatal', 'trace', 'debug', 'error', 'warn', 'info'].map(level => {
+    if (!logger[level]) {
+      fastifyLogger[level] = logLevels[level];
```
agerard-godaddy, line 20: You will want to fallback to the logger's methods, not the console's.

`debug`, `error`, `info`, and `warn` are guaranteed ([verify code](https://github.com/godaddy/gasket/blob/v7/packages/gasket-plugin-logger/lib/index.js#L15-L21)) to be on the logger.

```
const fallbackMap = {
  fatal: 'error',
  trace: 'debug'
}
```

```suggestion
      fastifyLogger[level] = logger[fallbackMap[level]];
```

bbetts-godaddy: Sounds good! Just made that update.

agerard-godaddy (approved):

PR #735, opened 2024-04-26
Added Additional JSDoc Typechecking

kbader-godaddy:
## Summary

Added typechecking to the following plugins:
- `@gasket/plugin-docusaurus`
- `@gasket/plugin-elastic-apm`
- `@gasket/plugin-fastify`
- `@gasket/gasket-utils`
- `@gasket/plugin-lint`

## Test Plan

All existing tests pass with the included `typecheck` scripts.


packages/gasket-utils/lib/apply-config-overrides.js
```diff
@@ -1,49 +1,71 @@
...
   );
 }
 
+/**
+ * Generator function to yield potential configurations
+ * @type {import('./index').getPotentialConfigs}
+ */
 function *getPotentialConfigs({ config, env, commandId, root, localFile }) {
   // Separate environment-specific config from other config
+  // @ts-ignore
```
agerard-godaddy, line 30: Why all the `@ts-ignores` in this function?

packages/gasket-utils/lib/package-manager.js
```diff
@@ -78,19 +70,19 @@ class PackageManager {
...
       const argv = [
-        '--loglevel', 'info',
-        '--spin', 'false',
-        '--progress', 'false'
+        '--loglevel',
+        'info',
+        '--spin',
+        'false',
+        '--progress',
+        'false'
```
agerard-godaddy, line 83: This was easier to understand with flag/value in same line

mmason2-godaddy (approved):

packages/gasket-utils/lib/package-manager.js
kbader-godaddy: Great feedback! Will update!

packages/gasket-utils/lib/apply-config-overrides.js
kbader-godaddy: This was included by mistake. Just pushed the update.

PR #742, opened 2024-05-07
[PFX-590] Move plugin engine to @gasket/core

jpina1-godaddy:
## Summary

To consolidate packages and type declarations, we will move the plugin engine to the new core package, and drop the standalone engine package.

## Changelog

- Moved PluginEngine to `@gasket/core`
- Added exports to `@gasket/core` to export cjs version of the package
- Removed `@gasket/engine`
- Update plugin deps and types to declare `@gasket/core`


packages/gasket-core/lib/index.d.ts
```diff
@@ -1,21 +1,115 @@
...
+  // To be extended by plugins
+  export interface HookExecTypes {}
 
-  export interface Gasket {
-    new (config: GasketConfigDefinition): Gasket
-    actions: GasketActions
-  }
 
-  export interface HookExecTypes {
+  export interface CoreHookExecTypes {
```
agerard-godaddy, line 12: I believe this needs stay `HookExecTypes` in order for `init`, etc to be available has `HookHandler`s
```suggestion
  // To be extended by plugins
  export interface HookExecTypes {
```

You can get rid of the empty declaration above it

jpina1-godaddy: @agerard-godaddy when I remove the empty declaration 10+ packages fail their typecheck. Would it be ok to leave this as is and create a new ticket to fix the types?

agerard-godaddy: I pushed some fixes per slack

packages/gasket-core/package.json
```diff
@@ -2,12 +2,19 @@
...
+    "import": "./lib/index.js",
+    "require": "./cjs/lib/index.js"
+  },
   "types": "lib/index.d.ts",
   "files": [
     "lib"
   ],
   "scripts": {
+    "clean": "rimraf cjs",
+    "build": "npm run clean && swc lib -d cjs",
```
agerard-godaddy, line 17: Another thing I just noticed, you can remove the extra clean script and rimraf dep. Instead do:
```suggestion
    "build": "swc lib -d cjs --delete-dir-on-start",
```

agerard-godaddy: Looks like using `--strip-leading-paths` is what we want to avoid the nested `lib` dir

packages/gasket-core/package.json
```diff
@@ -2,12 +2,20 @@
...
+    "require": "./cjs/index.js"
+  },
   "types": "lib/index.d.ts",
   "files": [
-    "lib"
+    "lib",
+    "cjs"
   ],
   "scripts": {
+    "clean": "rimraf cjs",
```
agerard-godaddy, line 17: not needed
```suggestion
```

packages/gasket-core/package.json
```diff
@@ -54,9 +64,13 @@
     "eslint-plugin-json": "^3.1.0",
     "eslint-plugin-unicorn": "^44.0.0",
     "jest": "^29.7.0",
+    "rimraf": "^3.0.2",
```
agerard-godaddy, line 67: not needed
```suggestion
```

agerard-godaddy (approved):
Couple small tweaks then 👍

mmason2-godaddy (approved):

PR #744, opened 2024-05-09
Ensure gasket/core is not bundled to allow GASKET_ENV

agerard-godaddy:
## Summary

With gasket files, being imported into app code, `@gasket/core` was being bundled. When this happens, any process.env vars are not captured, including `GASKET_ENV` which is necessary for configuring gaskets to be run in different environments.

Without this change, it would be necessary to `next build` uniquely for each environment.

## Changelog

**@gasket/plugin-nextjs**
- Add Webpack externals for `@gasket/core` in Next.js apps.

## Test Plan

N/A

### DIFF
--- a/packages/gasket-plugin-nextjs/lib/webpack-config.js
+++ b/packages/gasket-plugin-nextjs/lib/webpack-config.js
@@ -1,5 +1,13 @@
 const isGasketCore = /@gasket[/\\](core)/;
 
+/**
+ * Function to validate that '@gasket/core' is not used in browser code.
+ * If '@gasket/core' is found in the module request, an error is thrown.
+ *
+ * @param {Object} ctx - The context object containing the request string.
+ * @param {Function} callback - The externals callback.
+ * @returns {void}
+ */
 function validateNoGasketCore(ctx, callback) {
   if (isGasketCore.test(ctx.request)) {
     return callback(new Error('@gasket/core should not be used in browser code.'));
@@ -12,9 +20,9 @@ function validateNoGasketCore(ctx, callback) {
  * We do this to enable GASKET_ENV to be picked up and passed to the plugins.
  * Otherwise, unique builds per environment would be required.
  *
- * @param ctx
- * @param callback
- * @returns {*}
+ * @param {Object} ctx - The context object containing the request string.
+ * @param {Function} callback - The externals callback.
+ * @returns {void|*}
  */
 function externalizeGasketCore(ctx, callback) {
   if (isGasketCore.test(ctx.request)) {
--- a/packages/gasket-plugin-nextjs/lib/webpack-config.js
+++ b/packages/gasket-plugin-nextjs/lib/webpack-config.js
@@ -1,0 +1,57 @@
+const isGasketCore = /@gasket[/\\](core)/;
+
+function validateNoGasketCore(ctx, callback) {
+  if (isGasketCore.test(ctx.request)) {
+    return callback(new Error('@gasket/core should not be used in browser code.'));
+  }
+  return callback();
+}
+
+/**
+ * Externalize @gasket/core in the server build.
+ * We do this to enable GASKET_ENV to be picked up and passed to the plugins.
+ * Otherwise, unique builds per environment would be required.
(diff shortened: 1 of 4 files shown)


packages/gasket-plugin-nextjs/lib/webpack-config.js
```diff
@@ -0,0 +1,57 @@
+const isGasketCore = /@gasket[/\\](core)/;
+
+function validateNoGasketCore(ctx, callback) {
```
mmason2-godaddy, line 3: Nit: Doc block here.

mmason2-godaddy (approved):

packages/gasket-plugin-nextjs/lib/webpack-config.js
```diff
@@ -0,0 +1,57 @@
+const isGasketCore = /@gasket[/\\](core)/;
```
jpage-godaddy, line 1:
```suggestion
const isGasketCore = /@gasket[/\\]core[$/\\]/;
```

packages/gasket-plugin-nextjs/lib/webpack-config.js
```diff
@@ -0,0 +1,57 @@
...
+function validateNoGasketCore(ctx, callback) {
+  if (isGasketCore.test(ctx.request)) {
+    return callback(new Error('@gasket/core should not be used in browser code.'));
+  }
+  return callback();
+}
+
+/**
+ * Externalize @gasket/core in the server build.
+ * We do this to enable GASKET_ENV to be picked up and passed to the plugins.
```
jpage-godaddy, line 12: I just wonder if there's a way we can avoid using an environment variable altogether except in the case of initialization.

packages/gasket-plugin-nextjs/lib/webpack-config.js
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Configure Next.js Webpack for @gasket/core
+ *
+ * @param {Object} gasket - The gasket API.
+ * @param {Object} webpackConfig - Initial Next.js webpack config
+ * @returns {Object} Partial webpack config with UXCore2 support.
+ * @public
+ */
+function webpackConfigHook(gasket, webpackConfig) {
+  if (webpackConfig.name === 'client') {
```
jpage-godaddy, line 36: Is `name` the best way to detect this? Shouldn't we use the `isServer` property [from the docs](https://nextjs.org/docs/app/api-reference/next-config-js/webpack)?

packages/gasket-plugin-nextjs/lib/webpack-config.js
```diff
@@ -0,0 +1,57 @@
...
+    webpackConfig.externals.unshift(validateNoGasketCore);
+  } else {
+    webpackConfig.externals.unshift(externalizeGasketCore);
```
jpage-godaddy, line 39: A bit more succinct?

```suggestion
  webpackConfig.externals.unshift(webpackConfig.isServer ? externalizeGasketCore : validateNoGasketCore);
```

packages/gasket-plugin-nextjs/lib/webpack-config.js
agerard-godaddy: I'll try it out - we want it to externalize gasket/core from edge runtime bundles also, which is another non-client named config. If `isServer` covers that then 👍

packages/gasket-plugin-nextjs/lib/webpack-config.js
agerard-godaddy: As we move away from the CLI-driven approach, using this env var will be more necessary.

packages/gasket-plugin-nextjs/lib/webpack-config.js
agerard-godaddy: opting for the longer form as it is more readable and flexible, should we need to adjust the webpack config further.

jpina1-godaddy:
There's a ts test failure

jpina1-godaddy (approved):

PR #751, opened 2024-05-13
[PFX-557]: Remove @gasket/resolve

bbetts-godaddy:
## Summary

Ticket: [PFX-557](https://godaddy-corp.atlassian.net/browse/PFX-557)

We are removing the `@gasket/resolve` package to avoid the maintenance moving forward. 

## Changelog

- Remove `@gasket/resolve` package
- Remove references to `@gasket/resolve`
  
## Test Plan

Screenshot 2024-05-13 at 2 46 15 PM


packages/gasket-plugin-metadata/lib/utils.js
```diff
@@ -60,7 +55,6 @@ function safeAssign(target, other) {
 function loadAppModules(loader, app, modules) {
   app.modules = app.modules || [];
   Object.keys(app.package.dependencies)
-    .filter(name => !isPlugin(name) && !isPreset(name))
```
agerard-godaddy, line 63: Until we fully re-work this plugin, we should continue to filter out plugins and presets in this step. You could do so with a basic regexp like this:
```js
const rePluginPresets = /@gasket.+(plugin|preset)/;

// -- snipped --
  Object.keys(app.package.dependencies)
    .filter(name => !rePluginPresets.test(name))
```

packages/gasket-plugin-metadata/test/utils.test.js
```diff
@@ -122,14 +122,6 @@ describe('Utils', () => {
       expect(mockModules[0]).toHaveProperty('range', '^1.2.3');
       expect(mockModules[0]).toHaveProperty('from', 'my-app');
     });
-
-    it('ignores plugins and presets', () => {
```
agerard-godaddy, line 126: still a valid test for now

agerard-godaddy (commented):
Looking good, just one place where the logic should basically be intact for now.

mmason2-godaddy (approved):

agerard-godaddy (approved):

PR #757, opened 2024-05-16
Plugin default exports

agerard-godaddy:
## Summary

It is necessary to add default exports to satisfy Plugin import type checks.

## Changelog

- Added plugin default exports to index.d.ts
- Aligned plugin package main and types
- Removed unused conflicting metrics exec from Webpack plugin

## Test Plan

N/A


mmason2-godaddy (approved):
Noiceeeee

packages/gasket-plugin-docs-graphs/lib/index.js
kbader-godaddy, line 1: Curious about this change as most of the other plugins are using the `plugin` naming.

kbader-godaddy (approved):

packages/gasket-plugin-docs-graphs/lib/index.js
agerard-godaddy: Oh good eye. My IDE probably did it when I moved the filename to be aligned to index. I'll revert the variable name.

PR #758, opened 2024-05-17
[PFX-616] - @gasket/plugin-command

mmason2-godaddy:
## Summary
New Gasket v7 implementation for custom commands.

- [Plugin Readme](https://github.com/godaddy/gasket/tree/PFX-616/packages/gasket-plugin-command#gasketplugin-command)
- [Plugin hooks](https://github.com/godaddy/gasket/blob/PFX-616/packages/gasket-plugin-command/lib/index.js#L10)
- [Configure hook](https://github.com/godaddy/gasket/blob/PFX-616/packages/gasket-plugin-command/lib/configure.js) evaluates the `argv` and executes the `commands` lifecycle if `gasket` is found.
- [Simplified CLI implementation](https://github.com/godaddy/gasket/blob/PFX-616/packages/gasket-plugin-command/lib/cli.js)
- [Utils lifted from the CLI](https://github.com/godaddy/gasket/tree/PFX-616/packages/gasket-plugin-command/lib/utils)
- There's two `@ts-expect-error`(s) due to `A spread argument must either have a tuple type or be passed to a rest parameter.`. I'm open to suggestions 🤦‍♂️ 

### Follow-up Ticket
- Remove `@gasket/cli`
- Remove`@gasket/cli` from plugin dependencies

### Local Testing
```bash
npm i @gasket/plugin-command@7.1.1-canary.1
```

```js
// Create an inline plugin or a plugins in separate file
{
  name: 'test-plugin',
  hooks: {
    commands() {
      return {
        id: 'test-plugin',
        description: 'Test plugin',
        action: async () => {
          console.log('Custom Command!');
        }
      }
    }
  }
}
```

```bash
node ./gasket.js test-plugin
# Custom Command!
```

## Changelog
- Update package with version 7 custom commands implementation

## Test Plan
Tests, types and typechecking added.


lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "7.0.0-next.25",
+  "version": "7.1.1-canary.1",
```
agerard-godaddy, line 2: Can you back out this canary change?

```suggestion
  "version": "7.0.0-next.25",
```

packages/gasket-plugin-command/lib/configure.js
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable no-unused-vars, no-sync */
+import { gasketBin, processCommand } from './cli.js';
+const isGasketCommand = /\/gasket\.(js|ts)$/;
```
agerard-godaddy, line 3: We probably need to support mjs (and maybe cjs) extensions, also.

packages/gasket-plugin-command/README.md
```diff
@@ -0,0 +1,201 @@
...
+        action: async () => {
+          console.log('test-plugin');
+        }
+      }
+    }
+  }
+};
+```
+
+### Command Configuration
```
agerard-godaddy, line 50: I don't think you need this heading

packages/gasket-plugin-command/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+Commands can be configured with options(flags), arugments, defaults, and parsing functions. The type definitions for the command configuration are also available.
+
+#### Arguments
```
agerard-godaddy, line 54: heading tweak for consistency ([docs](https://github.com/godaddy/gasket/blob/main/CONTRIBUTING.md#markdown-documentation))
```suggestion
#### Example with Arguments
```

packages/gasket-plugin-command/README.md
```diff
@@ -0,0 +1,201 @@
...
+CLI. Executes the `commands` lifecycle in the `configure` hook of the
+`@gasket/plugin-command` plugin. This plugin utilizes [Commander.js] and additional documentation can be found in the documentation.
+
+## Installation
+
+```bash
+npm i @gasket/plugin-command
+```
+
+Update your `gasket.js` plugin configuration:
```
jpage-godaddy, line 13: Is this something developers will typically need to add to their app? I'm guessing it'll usually be included in presets, so maybe a note would be handy for that. Overall, I see an anti-pattern for Gasket users where they're frequently redundantly installing plugins.

agerard-godaddy (approved):
A few small tweaks then 👍🏻

packages/gasket-plugin-command/README.md
```diff
@@ -0,0 +1,201 @@
...
+          console.log('test-plugin');
+        }
+      }
+    }
+  }
+};
+```
+
+### Command Configuration
+
```
jpage-godaddy, line 51: It's weird that there's another third-level header here; seems like the content in this section could just be part of the `Commands` lifecycle.

```suggestion
```

packages/gasket-plugin-command/README.md
agerard-godaddy: It will typically be adding during create-gasket-app with a preset.

mmason2-godaddy: This is completely optional plugin and will NOT be included in any presets which is the opposite of version 6. Due to the foundational changes of version 7, apps will not be reliant on the Gasket CLI and or executing scripts with the CLI, which leads to this plugin being completely optional and intended to be used if the user needs a custom command option.

packages/gasket-plugin-command/README.md
```diff
@@ -0,0 +1,201 @@
...
+Commands can be configured with options(flags), arugments, defaults, and parsing functions. The type definitions for the command configuration are also available.
```
jpage-godaddy, line 52: I'm not sure how "The type definitions... are also available" is useful to developers without details on how to use the type definitions. I'd suggest the examples show how to use JSDocs to validate your hook.

packages/gasket-plugin-command/lib/index.d.ts
```diff
@@ -0,0 +1,107 @@
...
+  /* Command action handler */
+  action: (...args: any[]) => MaybeAsync<void>;
+
+  /* Hide from help output */
+  hidden?: boolean;
+
+  /* Default command to run if no command is provided */
+  default?: boolean;
+}
+
```
jpage-godaddy, line 27: Had a thought a while ago. You know how in the [Validating lifecycle scripts](https://github.com/godaddy/gasket/blob/main/docs/typescript.md#validating-lifecycle-scripts) document how you have to deal with annoyances of JSDocs syntax in order to use generics? Maybe it'd be a good pattern for plugins to export an easy-to-use type for hooks.

```typescript
// add Hook to import from `@gasket/core`

export type CommandHook = Hook<'command'>;
```

That way someone could slap this on their hook for validation:

```javascript
/**
* @type {import('@gasket/plugin-command').CommandHook}
 */
 configure(gasket) {
   // ...
 }
 ```

File: packages/gasket-plugin-command/README.md (hunk 7 above)
Comment (mmason2-godaddy) [reply packages/gasket-plugin-command/README.md]: The idea behind the line is simply to note that the configuration definition is available. Essentially noting the user can view the types for the complete scope of the configuration.

File: packages/gasket-plugin-command/lib/index.d.ts (hunk 8 above)
Comment (mmason2-godaddy) [reply packages/gasket-plugin-command/lib/index.d.ts:27]: So it would look like:

```js
export default {
  name: 'my-custom-command-plugin',
  hooks: {
    /* @type {import('@gasket/plugin-command').CommandHook} */
    commands() {
      return {
        // custom command
      }
    }
  }
}
```
?

packages/gasket-plugin-command/lib/utils/process-command.js
```diff
@@ -0,0 +1,50 @@
...
+  if (!isValidCommand(command)) throw new Error('Invalid command configuration');
+  const { id, description, action, args, options, hidden = false, default: isDefault = false } = command;
+  const cmd = program
+    .command(id)
+    .description(description)
+    .action(action);
+
+  if (args) {
+    const cmdArgs = processArgs(args);
+    // @ts-expect-error
```
jpage-godaddy, line 37: Hm, I'm not sure that `Array<string | boolean>` is a useful return type for `processArgs`; you should use a type where you don't have to expect a TypeScript error.

mmason2-godaddy: That is the accurate representation of the return type. Any suggestions on the typing?

mmason2-godaddy: Quite frankly that error is not helpful whatsoever. I've spent a lot of time trying to find a solution to no avail hence the `@ts-expect-error`.

packages/gasket-plugin-command/lib/index.d.ts
mmason2-godaddy: I've added the type export and a doc section in the [README](https://github.com/godaddy/gasket/tree/PFX-616/packages/gasket-plugin-command#example-with-jsdoc-types)

packages/gasket-plugin-command/lib/index.d.ts
```diff
@@ -0,0 +1,107 @@
...
+  /* Default value for the argument - never if required is true*/
+  default?: never;
+} | {
+  name: string;
+  description: string;
+  required?: false;
+  default?: any;
+};
+
+export type GasketCommandArg = Array<string | boolean>;
```
jpage-godaddy, line 54: Am I right in understanding that this is a type just used internally? I'd try to ensure that only types that the consuming developer cares about are exported.

packages/gasket-plugin-command/lib/utils/process-command.js
jpage-godaddy: You could be very specific and use `Parameters<Command['argument']>` if you want to explicitly make a type that is compatible with the parameters of a function.

packages/gasket-plugin-command/lib/index.d.ts
```diff
@@ -0,0 +1,109 @@
...
+
+export interface GasketCommandDefinition {
+  /* Command id/name */
+  id: string;
+
+  /* Command description */
+  description: string;
+
+  /* Command arguments */
+  args?: Array<GasketArgDefinition>;
```
jpage-godaddy, line 13: Any thoughts on this being an object instead? Personally I would find this more ergonomic:

```typescript
```suggestion
  args?: Record<string, GasketArgDefinition>;
```

...with `name` being implied by the key of the object. It should make it easier to "index" the arguments visually. I'd rather write:

```
args: {
  username: { },
  password: { }
}
```

...over:

```
args: [
  { name: 'username', ... },
  { name: 'password', ... }
]
```

mmason2-godaddy: It easily could be..... @agerard-godaddy thoughts?

mmason2-godaddy: Variadic options would be a little weird. If it's agreed I'll likely create a follow-up ticket.

```
args: {
  'no-prompt': { }
}
```

agerard-godaddy: 👍 I like the object suggestion

mmason2-godaddy: Ticket -> https://godaddy-corp.atlassian.net/browse/PFX-621?atlOrigin=eyJpIjoiYWQ4ZDU3OTBmMzEyNDE5MjhlYjM2M2E5MDdmNWUxNTIiLCJwIjoiaiJ9

kbader-godaddy (approved):

PR #762, opened 2024-05-21
Pfx 603

jpina1-godaddy:


mmason2-godaddy (approved):

packages/gasket-plugin-nextjs/lib/prompt.js
```diff
@@ -49,5 +49,18 @@ module.exports = async function promptHook(gasket, context, { prompt }) {
     newContext.addSitemap = addSitemap;
   }
 
+  if (!('useRedux' in context)) {
```
agerard-godaddy, line 52: These prompts should be in the redux plugin, not the nextjs plugin, right?

mmason2-godaddy: Depends.... they could be either place and the impact would if the plugin is in the preset dependencies or not essentially. If we add them to the redux plugin itself then the plugin will need to be a dependency of the preset(s).

agerard-godaddy: > If we add them to the redux plugin itself then the plugin will need to be a dependency of the preset(s)
Yes, I think that makes most sense. If it's not, selecting yes if the Nextjs plugin asks about it will have no effect.

agerard-godaddy: If a preset does not want redux to be an option, then it just does not include the redux plugin.

jpina1-godaddy: Moved prompt to the Redux plugin

packages/gasket-plugin-redux/lib/create.js
```diff
@@ -4,6 +4,8 @@ const { devDependencies } = require('../package.json');
 
 /** @type {import('@gasket/core').HookHandler<'create'>} */
 module.exports = async function create(gasket, context) {
+  if (Object.keys(context).includes('useRedux') && !context.useRedux) return;
```
agerard-godaddy, line 7: No sure that check is accurate. If `useRedux` is not on context, this check will fail and redux will be added.

More concise and accurate would be:
```suggestion
  if (context.useRedux !== true) return;
```

That is, if it's anything other than true, do not create with Redux

agerard-godaddy (approved):
small tweak then 👍

PR #768, opened 2024-05-24
Pfx 598

jpina1-godaddy:


mmason2-godaddy:
So this is just a utility that could be applied to the `@ux` packages?

jpina1-godaddy:
> So this is just a utility that could be applied to the `@ux` packages?

Yeah, the ticket said to reuse warn-if-outdated but the ux package really on needs to check the latest version of a package and doesn't need the caching or logging. I took out the check latest version logic and moved it to its own function to reuse in the ux package.

mmason2-godaddy (approved):

PR #770, opened 2024-05-24
Add missing type for version check

jpina1-godaddy:
## Summary

Follow up to this ticket PR https://github.com/godaddy/gasket/pull/768. I forgot to add its type to index.d.ts

### DIFF
--- a/packages/gasket-utils/lib/index.d.ts
+++ b/packages/gasket-utils/lib/index.d.ts
@@ -154,4 +154,4 @@ export interface ConfigContext {
 
 export function warnIfOutdated(pkgName: string, currentVersion: string): MaybeAsync<void>;
 
-export function getPackageLatestVersion(pkgName: string): Promise<string>;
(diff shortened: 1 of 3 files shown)


agerard-godaddy (approved):

packages/gasket-utils/lib/index.d.ts
```diff
@@ -153,3 +153,5 @@ export interface ConfigContext {
 
 
 export function warnIfOutdated(pkgName: string, currentVersion: string): MaybeAsync<void>;
+
+export function getPackageLatestVersion(pkgName: string): Promise<string>;
```
agerard-godaddy, line 157: Maybe also import the type to JSDocs [here](https://github.com/godaddy/gasket/blob/add-missing-type-for-version-check/packages/gasket-utils/lib/get-package-latest-version.js#L3-L8)

mmason2-godaddy (approved):

PR #773, opened 2024-06-03
Refactor data using GasketActions as @gasket/plugin-data

agerard-godaddy:
## Summary

We are moving away from response attachments and towards using GasketActions. We are also moving away from magic files. The data definition should be configured with the `makeGasket` config, meaning imports will be directly from the `gasket.(js|ts)` file.

## Changelog

**@gasket/plugin-data**
- Renamed plugin
- Added getGasketData and getPublicGasketData actions
- Lifecycles are now `gasketData` and `publicGasketData`
- No magic file imports

## Test Plan

N/A


mmason2-godaddy (approved):

packages/gasket-plugin-data/lib/create.js
```diff
@@ -0,0 +1,24 @@
+/// <reference types="create-gasket-app" />
+
+const { name, version, devDependencies } = require('../package.json');
+
+/**
+ * @type {import('@gasket/core').HookHandler<'create'>}
+ */
+async function create(gasket, { pkg, files, gasketConfig }) {
```
jpina1-godaddy, line 8: Does `gasketConfig.addPlugin` need to be called here to add the plugin to the app or is that not needed here?

jpina1-godaddy (approved):
Response data is still included here
https://github.com/godaddy/gasket/blob/6d9c80a0510deb988291719ff4fb704c94fb472c/packages/gasket-preset-api/lib/index.d.ts#L1

packages/gasket-plugin-data/lib/create.js
agerard-godaddy: Yes good catch!

PR #777, opened 2024-06-05
[PFX-637] Support create App Router

jpina1-godaddy:
## Summary

Added prompt to use App Router with the Gasket Nextjs plugin.

It does not currently work with Redux. This needs to be addressed in a separate ticket. We also need to port any logic from `_app` and `_document` files in a separate ticket.

I also removed the generated test from the cypress plugin. This test was conflicting with the generated cypress test in the nextjs plugin.

## Changelog

`@gasket/plugin-cypress`
- Removed generated index test

`@gasket/plugin-nextjs`
- Added App Router prompt
- Modified generated files to add App Router generated files

## Test Plan

- Unit tests
- Tested locally


packages/gasket-plugin-nextjs/generator/app/pages-router/pages/_document.js
mmason2-godaddy, line 1: Can you update with the latest from my PR that fixed Page Router rendering.

mmason2-godaddy (approved):

packages/gasket-plugin-nextjs/generator/app/pages-router/pages/_document.js
jpina1-godaddy: I believe this is the latest, right? https://github.com/godaddy/gasket/blob/030f993ec923aca546b17a4d9f875277f37490b1/packages/gasket-plugin-nextjs/generator/app/pages-router/pages/_document.js

packages/gasket-plugin-nextjs/generator/app/app-router/app/layout.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function RootLayout({ children }) {
+  return (
+    <html lang='en'>
```
jpage-godaddy, line 6: We shouldn't assume `lang='en'`

packages/gasket-plugin-nextjs/lib/prompt.js
```diff
@@ -3,8 +3,23 @@
...
+      {
+        name: 'useAppRouter',
+        message: 'Do you want to use the App Router? (experimental)',
+        type: 'confirm',
+        default: false
+      }
+    ]);
+
+    newContext.useAppRouter = useAppRouter;
+    newContext.nextServerType = 'defaultServer';
```
jpage-godaddy, line 18: Wouldn't this line get skipped if `useAppRouter` was already in the context?

packages/gasket-plugin-nextjs/generator/app/app-router/app/layout.js
jpina1-godaddy: Eslint looks for a value here `<html> elements must have the lang prop  jsx-a11y/html-has-lang`. 

What would be a good default for this? 

Would `en` be safe for the boilerplate code?

packages/gasket-plugin-nextjs/lib/prompt.js
jpina1-godaddy: I added a check outside of the useAppRouter prompt
```
  if (newContext.useAppRouter) {
    newContext.nextServerType = 'defaultServer';
  }
```

bbetts-godaddy (approved):

Issue #776, opened 2024-06-05
Allow Parallel Fetching of Multiple Locale Paths


eliebman-godaddy:
## Problem
We have an application that waits for i18n to be fetched for multiple dependencies.
Nesting multiple withLocaleRequired HOCs causes these locale files to be fetched in series, which means longer load times.
Each `withLocaleRequired` blocks the subsequent one from rendering until it's own fetching is complete.

Example Code:
```
// compose is a utility that wraps a component in multiple HOCs

export default compose(
  withLocaleRequired('/i18n'),
  withLocaleRequired('/i18n/modules/module-one'),
  withLocaleRequired('/i18n/modules/module-two'),
  withLocaleRequired('/i18n/modules/module-three'),
)(Component);
```
## Proposed Solution
Allow withLocaleRequired to accept an array of paths, to be fetched in parallel.
The only component blocked from rendering will be the underlying Component, and all locale fetching can take place simultaneously within one single instance of `withLocaleRequired`.

Example Code:
```
const requiredLocalePaths = [
  '/i18n',
  '/i18n/modules/module-one',
  '/i18n/modules/module-two',
  '/i18n/modules/module-three'
]

export default withLocaleRequired(requiredLocalePaths)(Component);
```

## Conclusion
A design that allows parallel fetching of many locale paths would speed up load times for our application significantly. I'm willing to work on this feature, but wanted to get feedback on the proposal first.

PR #779, opened 2024-06-06
Accept an array as `localesPath` in `useLocaleRequired`, `withLocaleRequired`, and `LocaleRequired`

eliebman-godaddy:
## Summary
Support passing an array of localePaths to `useLocaleRequired`, `withLocaleRequired`, and `LocaleRequired`
This helps optimize load times by fetching paths in parallel. Previously, consumers had to nest `withLocaleRequired` HOCs in order to ensure fetching of multiple paths, which caused requests to run in series, creating longer page load times.

Addresses Issue #776

## Changelog
- Accept an array as `localesPath` in `useLocaleRequired`, `withLocaleRequired`, and `LocaleRequired`

## Test Plan
Unit Tests are passing
Have not been able to verify inside an application due to node versioning issues, but we should probably verify this functionality with npm link before merging.

### DIFF
--- a/packages/gasket-react-intl/src/index.d.ts
+++ b/packages/gasket-react-intl/src/index.d.ts
@@ -38,7 +38,7 @@ export type LocaleRequiredWrapper = (props: {
  */
 export function withLocaleRequired<Props>(
   /** Path containing locale files */
-  localePathPart?: LocalePathPartOrThunk,
+  localePathPart?: LocalePathPartOrThunk | LocalePathPartOrThunk[],
   options?: {
     /** Custom component to show while loading */
     loading?: React.ReactNode;
@@ -50,7 +50,7 @@ export function withLocaleRequired<Props>(
 
 export interface LocaleRequiredProps {
   /** Path containing locale files */
-  localesPath: LocalePathPartOrThunk;
+  localesPath: LocalePathPartOrThunk | LocalePathPartOrThunk[];
   /** Custom component to show while loading */
   loading?: React.ReactNode;
 }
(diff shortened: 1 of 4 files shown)


packages/gasket-react-intl/src/index.d.ts
```diff
@@ -38,7 +38,7 @@ export type LocaleRequiredWrapper = (props: {
  */
 export function withLocaleRequired<Props>(
   /** Path containing locale files */
-  localePathPart?: LocalePathPartOrThunk,
+  localePathPart?: LocalePathPartOrThunk | LocalePathPartOrThunk[],
```
jpage-godaddy, line 41: If you're interested, there's a `MaybeMultiple` (or perhaps it was called `MaybeArray`) type that you can use.

jpage-godaddy:
You'll also need to update the server-side code for `withLocaleRequired` so it includes all of the strings in the server payload; currently it expects just a single thunk or string. You've only covered the client-side loading.

eliebman-godaddy:
> You'll also need to update the server-side code for `withLocaleRequired` so it includes all of the strings in the server payload; currently it expects just a single thunk or string. You've only covered the client-side loading.

@jpage-godaddy, It seems like serverside, in getInitialProps, we just [pass localePathPart directly to `serverLoadData`](https://github.com/eliebman-godaddy/gasket-ext/blob/c451b4ed9890019cce4b47a9cd03b6dfa10ce742/packages/gasket-react-intl/src/with-locale-required.js#L39), which looks like it [already supports arrays](https://github.com/eliebman-godaddy/gasket-ext/blob/c451b4ed9890019cce4b47a9cd03b6dfa10ce742/packages/gasket-helper-intl/lib/server.js#L12)

Is there some other server-side code I'm missing?

mmason2-godaddy (approved):

PR #785, opened 2024-06-11
[PFX-547]: Option for next custom server/fastify/express in presets

bbetts-godaddy:
## Summary

This PR allows for users to select which server type they would like to use with a preset. The options include `fastify`, `express` and `Custom Next Server`. 

Ticket: [PFX-547](https://godaddy-corp.atlassian.net/browse/PFX-547)

## Changelog

- Add prompt for server type to `@gasket/preset-api` and `@gasket/preset-nextjs`

## Test Plan

Screenshot 2024-06-11 at 10 42 21 AM


packages/gasket-preset-nextjs/lib/preset-prompt.js
```diff
@@ -18,4 +18,21 @@ export default async function presetPrompt(gasket, context, { prompt }) {
 
     Object.assign(context, { typescript });
   }
+
```
mmason2-godaddy, line 21: This should be in the nextjs plugin. You don't need the `customServer` option.

mmason2-godaddy (changes requested):

packages/gasket-plugin-nextjs/lib/index.d.ts
```diff
@@ -78,6 +78,7 @@ declare module 'create-gasket-app' {
     typescript: boolean;
     useRedux: boolean;
     useAppRouter: boolean;
+    server: 'express' | 'fastify' | 'customServer';
```
mmason2-godaddy, line 81:
```suggestion
    server: 'express' | 'fastify';
```

packages/gasket-preset-nextjs/test/preset-prompt.test.js
```diff
@@ -19,11 +19,51 @@ describe('presetPrompt', () => {
...
+  });
+
+  it('prompts for nextDevProxy when defaultServer', async () => {
+    mockContext.nextServerType = 'defaultServer';
+    await presetPrompt({}, mockContext, mockPrompt);
+    const first = mockPrompt.prompt.mock.calls[2][0][0];
+    expect(mockPrompt.prompt).toHaveBeenCalled();
+    expect(first.name).toEqual('nextDevProxy');
+  });
+
```
mmason2-godaddy, line 55: Maybe just add a test case for - `Does not prompt for server if nextServerType === defaultServer` then 👍

mmason2-godaddy (approved):

packages/gasket-preset-nextjs/lib/preset-prompt.js
```diff
@@ -1,3 +1,96 @@
+function prompts(context, prompt) {
```
jpina1-godaddy, line 1: Why were the Nextjs prompts moved to the preset-prompt? Would this lead to code duplication in the webapp preset and any other preset in the future that uses the next plugin.

mmason2-godaddy: They were moved because there was duplication. They should be removed from the plugin itself.

bbetts-godaddy: @jpina1-godaddy not sure if you saw this!

jpina1-godaddy: Where was the duplication? Weren't these only called in the nextjs plugin? 

If we want to use some of these prompts for the webapp preset, for example the promptAppRouter prompt, wouldn't we have to copy and paste this promptAppRouter prompt code from here into the webapp preset instead of sharing the code between the two presets if it was located in the nextjs plugin?

agerard-godaddy: Yea this change doesn't seem correct.

The general rule should be:
- Preset prompts present choices of which plugins to use.
- Plugin prompts present choices of which features to use.

If my present includes the a plugin, I expect that plugin to ask about what features to set up.
If I want to give my users options between different plugins, then I add that choice as a prompt to my preset. After that, if there are any feature choices, the plugin should ask about it.

So, unless I'm missing or overlooking something here, these Next.js prompts should remain in the Next.js plugin.

agerard-godaddy: Oh, it might be an ordering issue?

1. Preset prompts fire
2. Next plugin asks if you want Custom Server
3. if you do, then Preset can no longer as if you want Express or Fastify

Is that the scenario here?

agerard-godaddy: After a team chat, this is going to be explored further to define a good pattern moving forward. Approving for now to unblock.

agerard-godaddy (approved):

agerard-godaddy:
Closing in favor of https://github.com/godaddy/gasket/pull/811

PR #786, opened 2024-06-12
[PFX-578] - #3 Generate docs index

mmason2-godaddy:
## Summary
- Pull in `main` `site` & `scripts` changes
- [Update build process to generate index file](https://github.com/godaddy/gasket/blob/PFX-578-3/packages/gasket-assets/scripts/svg-to-react.js#L82-L87) in `@gasket/assets` - Reason: needed a proper `main` entry point in the package file.
- [Add check for plugin type and add a base case metadata hook to plugin if one doesn't exist](https://github.com/godaddy/gasket/blob/PFX-578-3/packages/gasket-core/lib/engine.js#L37-L52) - metadata hook if required if a plugin needs to end up in the docs - this also applies to preset or _"create only plugins"_.
- Cleanup missed `async configure` hooks
- [Tune metadata logic](https://github.com/godaddy/gasket/blob/PFX-578-3/packages/gasket-plugin-metadata/lib/actions.js#L36-L68)
- [Tune generate docs index file](https://github.com/godaddy/gasket/blob/PFX-578-3/scripts/generate-docs-index/index.js)
- [Add gasket js file](https://github.com/godaddy/gasket/blob/PFX-578-3/scripts/generate-docs-index/gasket.js) 

## Changelog
The above ☝️ 

## Test Plan
Tests updated


packages/gasket-plugin-metadata/lib/actions.js
```diff
@@ -28,23 +29,43 @@ module.exports = function actions(gasket) {
...
+          const pluginData = await handler(data);
+          pluginData.path = path.join(app.path, 'plugins');
+          plugins.push(pluginData);
+        } else {
+          const pluginData = await handler(data);
+          pluginData.path = path.dirname(path.join(require.resolve(pluginData.name), '..'));
+          const { dependencies, devDependencies } = require(path.join(pluginData.path, 'package.json'));
+
+          if (isGasketPreset)
+            presets.push(pluginData);
```
jpina1-godaddy, line 50: Would **@godaddy/preset-**  work properly here?

jpina1-godaddy (approved):

packages/gasket-plugin-metadata/lib/actions.js
mmason2-godaddy: It wouldn't however I'm not sure we should pollute the OS scope with `@godaddy`? This is a good point and I'm not sure.

bbetts-godaddy (approved):

PR #794, opened 2024-06-25
[PFX-623]: Tuneup: @gasket/plugin-happyfeet

bbetts-godaddy:
## Summary

Ticket: [PFX-623](https://godaddy-corp.atlassian.net/browse/PFX-623)

Update the HappyFeet plugin to use GasketActions and remove the preboot hook to align with v7 upgrades.

## Changelog

- remove preboot hook
- add GasketActions

## Test Plan

Screenshot 2024-06-25 at 11 03 15 AM

- Add new test suite for actions


packages/gasket-plugin-happyfeet/README.md
```diff
@@ -50,15 +50,52 @@ module.exports = {
...
+```
+
+```js
+const happy = gasket.actions.getHappyFeet();
+```
+
+#### happyConfig example
+
+```js
+const happyConfig =  {
```
jpina1-godaddy, line 87: Why have 2 different ways to add config to happy-feet? 

Wouldn't the same happy-feet instance be used throughout the app and therefore just look at the gasket config for configuration?

packages/gasket-plugin-happyfeet/lib/actions.js
```diff
@@ -0,0 +1,10 @@
+const happyFeet = require('happy-feet');
+
+/** @type {import('@gasket/core').HookHandler<'actions'>} */
+module.exports = function actions(gasket) {
+  return {
+    getHappyFeet(happyConfig) {
+      return happyFeet(gasket.config.happyFeet || happyConfig || {});
```
jpina1-godaddy, line 7: If you end up using the same instance for the whole app you should set a single instance to reference in the action file and then return that every time you call the action.

Something like below

```
let  happy = null

module.exports = function actions(gasket) {
 if(!happy) {
happy = happyFeet(gasket.config.happyFeet || {})
}
return happy
```

bbetts-godaddy: @jpina1-godaddy I had this originally, but changed it per t[his thread](https://godaddy.slack.com/archives/C034EH5GG4R/p1719248113151249?thread_ts=1719009070.161629&cid=C034EH5GG4R)

jpina1-godaddy: @mmason2-godaddy Wouldn't this code create a new instance of happy-feet every time you call it?  I'm not too familiar with happy feet but I think you would want to create the one instance and then reference it later instead of creating a new one each time. Kind of like how it was before when it was on the gasket object but instead of it living on the gasket object it would reference the top level variable in this file.

mmason2-godaddy: I'm not super familiar with this either. I'm good with that implementation 👍 . Sorry for the back and forth!

jpina1-godaddy: No worries, thanks for taking the time to send feedback.

mmason2-godaddy (approved):

jpina1-godaddy (approved):

PR #803, opened 2024-06-27
[PFX-625] - Add custom command to execute the build lifecycle

mmason2-godaddy:
## Summary
Instead of removing build hooks and or porting them to new lifecycles we've opted to keep the `build` lifecycle and utilize it via a custom command that can be executed in npm scripts.

## Changelog
- Add custom command to execute the build lifecycle

## Test Plan
Update tests

### DIFF
--- a/packages/gasket-plugin-command/lib/index.d.ts
+++ b/packages/gasket-plugin-command/lib/index.d.ts
@@ -99,7 +99,7 @@ export function processCommand(command: GasketCommandDefinition): GasketCommand;
 declare module '@gasket/core' {
 
   export interface HookExecTypes {
-
+    build(): any;
     commands(): GasketCommandDefinition;
   }
 }
--- a/packages/gasket-nextjs/lib/document/index.js
+++ b/packages/gasket-nextjs/lib/document/index.js
@@ -74,7 +74,7 @@ export function withGasketData(
   return Document => {
 
     async function getInitialProps(ctx) {
-      const gasketData = await gasket.actions.getPublicGasketData?.(ctx.req) ?? {};
+      const gasketData = ctx.req ? await gasket.actions.getPublicGasketData(ctx.req) : {};
 
       return {
         ...(await Document.getInitialProps(ctx)),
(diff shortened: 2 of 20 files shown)


packages/gasket-plugin-command/lib/index.d.ts
```diff
@@ -99,7 +99,7 @@ export function processCommand(command: GasketCommandDefinition): GasketCommand;
 declare module '@gasket/core' {
 
   export interface HookExecTypes {
-
+    build(): any;
```
mmason2-godaddy, line 102: Left as `any` due to the fact that I'm not 100% on the build hook signature's in the internal repo.

jpina1-godaddy (commented):
What do you think about moving the code that adds the prebuild script to the create-gasket-app package like in the [write-pkg.js](https://github.com/godaddy/gasket/blob/b62542e8dd2932e4f39a1940ef1d69625cbb3717/packages/create-gasket-app/lib/scaffold/actions/write-pkg.js#L11) file instead of adding a prebuild script for every plugin that needs it? Then it would be included for all gasket apps by default.

mmason2-godaddy:
> What do you think about moving the code that adds the prebuild script to the create-gasket-app package like in the [write-pkg.js](https://github.com/godaddy/gasket/blob/b62542e8dd2932e4f39a1940ef1d69625cbb3717/packages/create-gasket-app/lib/scaffold/actions/write-pkg.js#L11) file instead of adding a prebuild script for every plugin that needs it? Then it would be included for all gasket apps by default.

That's a interesting idea. The reality is the `prebuild` script is only necessary if you an app contains a plugin with a `build` hook. OS NextJS preset for example doesn't have any plugins with a `build` hook. That script would essentially do nothing which isn't great but isn't terrible. 

We only need to add the `prebuild` script addtion in the `create` hook in two plugins - `intl & swagger` - that covers all the presets.

jpina1-godaddy:
If it's not used that often I just adding them as need works for me 👍

jpina1-godaddy (approved):

kbader-godaddy (approved):

PR #334, opened 2022-02-03
Next 12 no babel config

kawikabader:
## Summary

New Gasket apps should not have a `.babelrc` generated by default. This opts them out of using SWC by Next.

## Changelog

Updates to remove babel configs.

## Test Plan

Tests updated.

PR #804, opened 2024-06-28
[PFX-664] Fix testing suites in v7

jpina1-godaddy:
## Summary

Fix test suites in v7 that fail when trying to run their default tests in app code.

## Changelog

**@gasket/plugin-cypress**
- update scripts and generated files

**@gasket/plugin-jest**
- update generated files

**@gasket/plugin-mocha**
- update scripts and generated files

**@gasket/plugin-mocha**
- update esm syntax and generated files

## Test Plan
- Unit tests
- Tested locally


mmason2-godaddy (approved):

packages/gasket-plugin-mocha/generator/.babelrc
```diff
@@ -0,0 +1,5 @@
+{
```
agerard-godaddy, line 1: Doesn't this cause the app to opt out of SWC? I believe that is why we moved to babel/register per https://github.com/godaddy/gasket/pull/334

jpina1-godaddy: I've seen a babel config in the linaria plugin so I thought it was ok here. There is configuration I can add to the nextConfig that will force the app to use swc here https://nextjs.org/docs/messages/swc-disabled#possible-ways-to-fix-it.

agerard-godaddy: Does it no longer work anymore with `@babel/register` in a setup script? Perhaps adjust it to what you have in the babelrc?
```
require('@babel/register')({
  presets: [
    '@babel/preset-react'
  ]
});
```

jpina1-godaddy: This doesn't work for me either. I'm not sure if babel register will help us now that we are using esm. There is a note at the bottom of the babel register docs talking about how it will not compile esm on the fly https://babeljs.io/docs/babel-register. 

> Note: @babel/register does not support compiling native Node.js ES modules on the fly, since currently there is no stable API for intercepting ES modules loading.

agerard-godaddy: So, you've brought in `@node-loader/babel` presumably to _compile all files with babel before they are executed in Node_. It doesn't feel great that in order to use Mocha for unit tests, building the app must opt-out of SWC..

Short of forking or contributing `@node-loader/babel` to accept babel options in a wrapper, and thus avoid a babel config file, I don't see any better options atm. I guess we can move forward with what you have, but please add a ticket to review approaches for Mocha support - maybe we just don't in v7 to avoid the messiness.

agerard-godaddy (approved):

packages/gasket-plugin-mocha/generator/.babelrc
jpina1-godaddy: Created a new ticket to address the mocha implementation concerns in the upcoming sprint

PR #806, opened 2024-07-03
Exclude gaskets from client bundling

agerard-godaddy:
## Summary

In order to allow gaskets to be imported for server-side `getInitialProps` usage cases, we need to ensure they are not attempted to be bundled for the browser.

## Changelog

**create-gasket-app**
- Default generate gaskets with filename configured

**@gasket/plugin-nextjs**
- Add false alias for gasket filenames for client bundling

## Test Plan

Locally tested by calling `/path/to/gasket/packages/create-gasket-app/lib/index.js`

### DIFF
--- a/packages/gasket-core/lib/index.d.ts
+++ b/packages/gasket-core/lib/index.d.ts
@@ -50,10 +50,11 @@ declare module '@gasket/core' {
 
   // This is the config
   export interface GasketConfig {
-    filename?: string;
     plugins: Array<Plugin>;
     root: string;
     env: string;
+    /** Path to the gasket instance file. Can be set to `import.meta.filename` **/
+    filename?: string;
   }
 
   export class GasketEngine {
--- a/packages/gasket-core/lib/index.d.ts
+++ b/packages/gasket-core/lib/index.d.ts
@@ -40,8 +40,8 @@ declare module '@gasket/core' {
 
   export type Plugin = {
     name: string;
-    version: string;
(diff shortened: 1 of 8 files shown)


packages/gasket-core/lib/index.d.ts
```diff
@@ -40,8 +40,8 @@ declare module '@gasket/core' {
 
   export type Plugin = {
     name: string;
-    version: string;
-    description: string;
+    version?: string;
+    description?: string;
```
mmason2-godaddy, line 44: If these are optional then the metadata will missing if they are not defined.

packages/gasket-core/lib/index.d.ts
```diff
@@ -50,6 +50,7 @@ declare module '@gasket/core' {
 
   // This is the config
   export interface GasketConfig {
+    filename?: string;
```
mmason2-godaddy, line 53: So you're declaring the name of the file in which this is defined? In `gasket.js` you'll have a `filename` property with the value of `gasket.js`?

agerard-godaddy: Basically yes, though it'll be the full path i.e. `/absolute/path/to/gasket.js`. By using `import.meta.filename` we don't have to worry if the user renames or moves it.

mmason2-godaddy (approved):

rxmarbles (approved):

PR #808, opened 2024-07-11
Fix locale files loading multiple times

rcastillogerman-godaddy:
## Summary

With the v6.46.1 release, a bug was introduced where locale files were loaded multiple times when using the `withLocaleRequired` high-order component. This was due to the removal of a default value for the `status` property in the `init` function for the `withIntlProvider` hook.

## Changelog

Provide a default value for the status value used in the GasketIntlContext component.

### DIFF
--- a/packages/gasket-react-intl/src/context.js
+++ b/packages/gasket-react-intl/src/context.js
@@ -4,7 +4,7 @@ import { manifest } from './config';
 /** @type {import('./index').GasketIntlContext} */
 const defaultContext = {
   locale: manifest.defaultLocale,
-  status: {},
+  status: {}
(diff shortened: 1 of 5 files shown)


agerard-godaddy (approved):

kbader-godaddy:
@rcastillogerman-godaddy Thanks for the contribution! Could we also get tests updated to cover this change?

sfenerin-godadddy (approved):

rcastillogerman-godaddy:
> @rcastillogerman-godaddy Thanks for the contribution! Could we also get tests updated to cover this change?

Addressed.

PR #810, opened 2024-07-15
[PFX-663] Fix Analyze Command

jpina1-godaddy:
## Summary

The @gasket/plugin-analyze was failing because it relied on the command property of the gasket object, which is no longer available in version 7. To resolve this issue, this PR uses an environment variable to trigger the bundle analysis code.

## Changelog

**@gasket/plugin-analyze**
- Remove the custom command and replace it with an npm script that sets an environment variable. The webpackConfig lifecycle in the plugin will fire when this variable is true.

## Test Plan
- Updated unit tests
- Tested locally


agerard-godaddy (approved):
🎉

mmason2-godaddy (approved):

PR #812, opened 2024-07-16
[PFX-671] - ESLint Typescript Support

mmason2-godaddy:
## Summary
ESLint Typescript Support
- Add file extensions to npm script based on the `context.typescript` value
- Adjust the character usage from the generated `gasket.ts|.js` file(remove tabs)
- Add `parser` support to eslint package.json config type

## Changelog
- ESLint Typescript Support

## Test Plan
Tests & types updated

### DIFF
--- a/packages/gasket-plugin-lint/lib/code-styles.js
+++ b/packages/gasket-plugin-lint/lib/code-styles.js
@@ -203,8 +203,10 @@ const common = {
     // Handle common eslint configuration
     if (hasEslint) {
       if (!pkg.has('scripts', 'lint')) {
+        const exts = `.js,.jsx,.cjs${typescript ? ',.ts,.tsx' : ''}`;
+
         pkg.add('scripts', {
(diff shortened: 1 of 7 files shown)


packages/gasket-plugin-lint/lib/code-styles.js
```diff
@@ -204,11 +204,21 @@ const common = {
     if (hasEslint) {
       if (!pkg.has('scripts', 'lint')) {
         pkg.add('scripts', {
-          'lint': 'eslint --ext .js,.jsx,.cjs .',
+          'lint': `eslint --ext .js,.jsx,.cjs${typescript ? ',.ts,.tsx' : ''} .`,
```
agerard-godaddy, line 207: nit: it would be a little easier to digest this if it were split up some.

```suggestion
          'lint': `eslint --ext ${ext} .`,
```

with `ext` being something like:
```js
const ext = '.js,.jsx,.cjs' + (typescript ? ',.ts,.tsx' : '')
```

agerard-godaddy (approved):
small nit, otherwise 👍

jpina1-godaddy (approved):

PR #816, opened 2024-07-24
[PFX-673] -  Tune up APM Plugin

mmason2-godaddy:
## Summary
- Remove `preboot`
- Add `getApmTransaction` action
- Refactor create
- Tune `apmTransaction` hook signature(drop the use of `res`)
- Tune types
- Adjust middleware

## Changelog
-  Tune up APM Plugin: Gasket Actions, remove preboot, adjust signatures, refactor hooks

## Test Plan
Tests & types updated


packages/gasket-core/lib/index.d.ts
```diff
@@ -1,3 +1,5 @@
+import { IncomingMessage } from 'node:http'
```
agerard-godaddy, line 1: Unused - from testing? You can probably revert the changes to this file

packages/gasket-plugin-elastic-apm/lib/actions.js
```diff
@@ -0,0 +1,21 @@
+/** @type {import('@gasket/core').HookHandler<'actions'>} */
+module.exports = function actions(gasket) {
+  return {
+    async getApmTransaction(req) {
+      const apm = gasket.apm;
```
agerard-godaddy, line 5: I'm not seeing how `gasket.apm` is ever assigned.

agerard-godaddy: You still have other places looking for `gasket.apm`, as well as some tests setting it.

packages/gasket-plugin-elastic-apm/lib/create.js
```diff
@@ -0,0 +1,15 @@
+const { devDependencies } = require('../package.json');
+
+/** @type {import('@gasket/core').HookHandler<'create'>} */
+module.exports = function create(gasket, { pkg, files }) {
+  const generatorDir = `${__dirname}/../generator`;
+
+  pkg.add('dependencies', {
```
jpina1-godaddy, line 7: Do you need to add the plugin to the gasket file and as a dependency to use the action?

```
  gasketConfig.addPlugin('pluginElasticApm', name);
  pkg.add('dependencies', {
    [name]: `^${version}`
  }); 
```

packages/gasket-plugin-elastic-apm/lib/actions.js
```diff
@@ -0,0 +1,21 @@
+/** @type {import('@gasket/core').HookHandler<'actions'>} */
+module.exports = function actions(gasket) {
+  return {
+    async getApmTransaction(req) {
```
jpina1-godaddy, line 4: I see there are some cases where the apm transaction will not be returned. I would think that this action would always return the apm transaction when you call it. I'm not at all familiar with the api though. Is it possible to return the apm transaction event if it has already started?

If not maybe a name like `startApmTransation ` could make more sense for this action?

jpina1-godaddy (commented):
Do you think the package is in a state where you could add some docs for the new action and remove docs for the preboot hook?

jpina1-godaddy (approved):

packages/gasket-plugin-elastic-apm/lib/create.js
```diff
@@ -0,0 +1,20 @@
...
+  gasketConfig.addPlugin('pluginElasticApm', name);
+
+  pkg.add('dependencies', {
+    [name]: `^${version}`,
+    'dotenv': devDependencies.dotenv,
+    'elastic-apm-node': devDependencies['elastic-apm-node']
+  });
+
+  pkg.add('scripts', {
+    start: 'gasket start --require ./setup.js'
```
agerard-godaddy, line 16: We don't have a gasket start anymore

packages/gasket-plugin-elastic-apm/lib/create.js
```diff
@@ -0,0 +1,24 @@
...
+  pkg.add('dependencies', {
+    [name]: `^${version}`,
+    'dotenv': devDependencies.dotenv,
+    'elastic-apm-node': devDependencies['elastic-apm-node']
+  });
+
+  pkg.extend((current) => {
+    return {
+      scripts: {
+        start: `node --import ./setup.js & ${current.scripts.start}`
```
agerard-godaddy, line 18: `&` is going to be a different process..

I think what you want here is to set [NODE_OPTIONS](https://nodejs.org/api/cli.html#-r---require-module) per examples in the [APM docs](https://www.elastic.co/guide/en/apm/agent/nodejs/current/nextjs.html#_step_2_start_the_apm_agent_2):
```suggestion
        start: `NODE_OPTIONS=--import=./setup.js ${current.scripts.start}`
```

Can you give that a shot? It should work with either Next or API start scripts

PR #827, opened 2024-08-01
[PFX-577] Docs Update: Move CLI docs

jpina1-godaddy:
## Summary

CLI docs exist in https://github.com/godaddy/gasket/tree/main/packages/gasket-cli/docs  that need a review and revival.

## Changelog

**docs**
- Added revived docs

**@gasket/core**
- Added gasket engine docs


docs/configuration.md
```diff
@@ -0,0 +1,140 @@
...
+  }
+}
+```
+
+The last step is making sure that Gasket is told which environment it is running
+in when starting up. There are three ways of doing this. The first would be to
+set the `GASKET_ENV` environment variable. The second would be to pass an `--env`
+command line argument when running any `gasket` command. The third is to
+programmatically set an `env` property at the top level of your configuration,
+which would enable you to use any arbitrary environment derivation logic.
```
agerard-godaddy, line 92: The `--env` is no longer an option. We don't support programmatically setting `env`, though perhaps we could?

Looking at [this line](https://github.com/godaddy/gasket/blob/v7/packages/gasket-core/lib/index.js#L54), perhaps we could do something like:
```
const env = gasketConfig.env ?? getEnvironment();
```

docs/configuration.md
```diff
@@ -0,0 +1,140 @@
...
+event. Your hook will be passed a config object, and you'll return an augmented
+config object or a Promise containing one for async config.
+
+```js
+// theoretical.js plugin
+export default {
+  name: 'theoretical',
+  hooks: {
+    async configure(gasket, baseConfig) {
+      const remoteConfig = await theoreticalClient.getConfig(baseConfig.theoretical.url);
```
agerard-godaddy, line 132: The `configure` lifecycle is not long async - can you change up this example?

docs/plugins.md
```diff
@@ -0,0 +1,215 @@
...
+```
+
+The optional `dependencies` array lists other plugins that must be installed as
+a prerequisite. The `hooks` map has a key for every event the plugin handles,
+with the values being either the function itself or an object specifying
+`timing` options and the `handler`.
+
+## Naming convention
+
+Plugins and presets should adhere to the project-type prefixed naming convention. This formatting allows user plugins to be referenced with short names and will help avoid collisions. This convention also mimics those by other projects such a `@babel` and `@oclif`
```
agerard-godaddy, line 36: The naming convention is not a hard requirement anymore since we import plugins. Can you rework this as a "recommended naming convention" with some examples?

docs/plugins.md
```diff
@@ -0,0 +1,215 @@
...
+        after: [ 'metadata' ]
+      },
+      handler: async function initHook(gasket) {
+        const { config } = gasket;
+        console.log(config.name);
+      }
+    }
+  }
+}
+```
```
agerard-godaddy, line 66: This is another outdated example. Also, we do not support "short plugin names" in the timing part. Can you spin up a different, more accurate example?

docs/plugins.md
```diff
@@ -0,0 +1,215 @@
...
+      return 'That subtle off-white coloring';
+    },
+    alibi: function () {
+      return 'Was returning some video tapes';
+    }
+  }
+}
+```
+
+Here are some basic tests, assuming we're using the `mocha` test framework.
```
agerard-godaddy, line 126: Can you adjust this example to jest since we've finally aligned the repo on it?

docs/plugins.md
```diff
@@ -0,0 +1,215 @@
...
+```
+
+Then, upon running `node gasket.js docs`, developers will automatically find
+documentation for the `detective` plugin.
+
+## One-off plugins
+
+While it is encouraged to build plugins as separate packages, the ability to
+create one-off plugins in an app is available. This gives you access to tie into lifecycles, set timings, or even
+add your own hooks.
```
agerard-godaddy, line 205: nit: formatting
```suggestion
While it is encouraged to build plugins as separate packages, the ability to
create one-off plugins in an app is available. This gives you access to tie
into lifecycles, set timings, or even add your own hooks.
```

docs/presets.md
```diff
@@ -0,0 +1,104 @@
...
+Much like [babel presets], Gasket presets allow common plugins to be grouped and
+loaded together. They serve 2 purposes: to serve as codified sets of plugins,
+and to facilitate rapid creation of Gasket application.
+
+At GoDaddy, we have presets specifically tailored with internal sets of plugins,
+making maintaining standards around authentication, style, analytics, and more
+significantly easier.
+
+See the [naming conventions] for how to best name a preset, ensuring that
+Gasket's plugin engine properly resolves it.
```
agerard-godaddy, line 14: This whole guide needs to be quite a bit different now. Presets are no longer like babel's, as they are only used to provide configuration and prompting to facilitate app generation using `create-gasket-app`.

If you want to take a stap at reworking this doc now, go for it. Otherwise, spin up a separate ticket for us to address as a follow up.

docs/configuration.md
jpina1-godaddy: Should this be in another ticket or is it ok to add to this PR?

docs/configuration.md
```diff
@@ -0,0 +1,140 @@
...
+  express: async function (gasket, express) {
+    console.log(gasket.config)
+  }
+};
+
+export default { name, hooks };
+
+// app-code.js
+
+import gasket from './gasket';
```
agerard-godaddy, line 112: Show the extension

```suggestion
import gasket from './gasket.js';
```

docs/plugins.md
```diff
@@ -0,0 +1,201 @@
...
+  }
+};
+```
+
+The optional `dependencies` array lists other plugins that must be installed as
+a prerequisite. The `hooks` map has a key for every event the plugin handles,
+with the values being either the function itself or an object specifying
+`timing` options and the `handler`.
+
+## Recommended naming convention
```
agerard-godaddy, line 33: Nice and neat 👍

agerard-godaddy (approved):
A couple small things then 👍

docs/configuration.md
```diff
@@ -0,0 +1,139 @@
...
+    "url": "https://your-datacenter1-prod-service-endpoint.com"
+  }
+}
+```
+
+The last step is making sure that Gasket is told which environment it is running
+in when starting up. There are three ways of doing this. The first would be to
+set the `GASKET_ENV` environment variable. Alternatively, you can
+programmatically set an `env` property at the top level of your configuration,
+enabling you to use any arbitrary environment derivation logic.
```
agerard-godaddy, line 91:
```suggestion
in when starting up. To do so, set the `GASKET_ENV` environment variable.
Alternatively, you can programmatically set an `env` property at the top level
of your configuration, enabling you to use any arbitrary environment derivation logic.
```

mmason2-godaddy (approved):

PR #840, opened 2024-08-07
Ignore range prefixes for prerelease versions

agerard-godaddy:
## Summary

When adding dependency version for packages during create, we need to ignore any range prefixes for prerelease versions as this has been seen to trip up npm when installing dependencies.

## Changelog

**create-gasket-app**
- ignore range prefixes for prerelease versions

## Test Plan

Tested locally by creating an app while invoke `/path/to/gasket/packages/create-gasket-app/lib/index.js`


mmason2-godaddy (approved):
Noiceee

bbetts-godaddy (approved):

PR #841, opened 2024-08-08
Error on recursive lifecycles

agerard-godaddy:
## Summary

With the introduction of Gasket Actions, it's much easier to accidentally end up in a recursive lifecycle situation, leading to a `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`.

This PR overhauls things for improved tracking across lifecycles to help avoid circular executions. Branch tracking and more granular tracing has been added to also help with debugging.

Traces are grouped by branch, and have category icons
- ⋌ branch start
- ★ action start
- ◆ sync lifecycle start
- ◇ async lifecycle start
- ↪ lifecycle hook start

To also trace actions execution and tracking, how actions are registered needed to be adjust and better coupled with the engine.

## Test Plan

Screenshot 2024-08-15 at 7 59 33 PM


kbader-godaddy (approved):

bbetts-godaddy (approved):

packages/gasket-core/lib/engine.js
```diff
@@ -12,7 +14,7 @@ class GasketEngine {
 
     this._hooks = {};
     this._plans = {};
-    this._traceDepth = 0;
+    this._traceStack = [];
```
agerard-godaddy, line 17: I am realizing that this isn't exactly safe.

```

const worker1 = gasket.exec('do-something-that-takes awhile');
const worker2 = gasket.exec('do-something-that-takes awhile');

await Promise.all([worker1, worker2]);
```

This will increase the trace depth. While not a common scenario, it is bound to occur when handling multiple requests.

I'll add some more tests for now and cycle back to this.

agerard-godaddy: Accounting for this scenario accurately ended up snowballing into a decent amount. I'm requesting re-reviews since of the PR as a result.

kbader-godaddy (approved):

mmason2-godaddy (approved):

PR #860, opened 2024-08-21
feat(plugin-express): Introduce getExpressApp action

scuilla:
## Summary

To enable the use of gasket actions and config in Express servers through direct imports, we need to restructure our import patterns. Instead of importing `{ routes } from './gasket'`, we now call app within the routes index. This results in a document  that more closely resembles a non-Gasket Express apps, hopefully making them easier to understand.

```js

const app = gasket.actions.getExpressApp();

app.get("/default", (req, res) => { 
  res.send("default route"); 
}); 
```

### `gasket-plugin-express`
* introduced new `getExpressApp` gasket action
* removed `routes` prop on express config
* removed types associated with the routes config
* updated generators with new import pattern

### `gasket-plugin-https`
* use `gasket.root.exec()` instead of `gasket.exec()`, which fixes an edge case where calling `/healthcheck` was flagging the recrusive action check introduced in https://github.com/godaddy/gasket/pull/853

### `gasket-plugin-typescript`
* updated generator with changes

### `gasket-preset-api`
* updated generator with changes

## Test Plan

* removed all tests that relied on the `routes` import pattern
* added test for typescript glob

### DIFF
--- a/packages/gasket-plugin-docs/lib/create.js
+++ b/packages/gasket-plugin-docs/lib/create.js
@@ -6,10 +6,24 @@ const { name, version } = require('../package.json');
 const { DEFAULT_CONFIG } = require('./utils/constants');
 
 /** @type {import('@gasket/core').HookHandler<'create'>} */
-module.exports = function create(gasket, { pkg, gasketConfig, gitignore }) {
+module.exports = function create(gasket, {
+  pkg,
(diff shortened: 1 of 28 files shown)


mmason2-godaddy (approved):

mmason2-godaddy:
🙌

PR #863, opened 2024-08-21
Simplify recursion protection methods

agerard-godaddy:
## Summary

Requiring explicit branching for executing repeated lifecycles (i.e. in a loop) was not good DX. Also, we don't really need to restrict executing actions multiple times for fear of recursions. In fact, it is expected that actions be called multiple times down a lifecycle chain.

## Changelog

- Allow multiple action calls
- Auto-isolate trace stacks for repeating lifecycle calls
- Add debugging docs ([rendered markdown](https://github.com/godaddy/gasket/blob/recursion-protection/packages/gasket-core/README.md))

## Test Plan

Screenshot 2024-08-21 at 2 59 28 PM


kbader-godaddy (approved):

kbader-godaddy:
LGTM 👍

mmason2-godaddy (approved):

PR #873, opened 2024-08-27
[PFX-725] Update Middleware Plugin Types

kbader-godaddy:
## Summary

There were a few small tweaks needed to better align the types definitions for the middleware plugin.

## Changelog

Small tweaks to middleware plugin types definitons.

## Test Plan

All unit and type-check tests pass.


packages/gasket-plugin-middleware/lib/index.d.ts
```diff
@@ -1,15 +1,73 @@
...
+} from 'http2';
+
+/** Type alias for Fastify application with HTTP/2 support */
+type FastifyApp<
+  Server = Http2SecureServer,
+  Request = Http2ServerRequest,
+  Response = Http2ServerResponse
+> = FastifyInstance<Server, Request, Response>;
+
+declare namespace Internal {
```
agerard-godaddy, line 17: Personally, I like the separation of internal types in a different file. It makes the jsdoc imports simpler and cleaner. Your change here seems to suggest the Internal namespace type will also be accessible outside the module as well, which seems not desirable. What does putting them all in the `index.d.ts` gain us?

packages/gasket-plugin-middleware/README.md
```diff
@@ -69,33 +86,41 @@ export default makeGasket({
...
   ]
+  ...
+});
+
 ```
 
 ## Lifecycles
 
-### middleware
+### `middleware`
```
agerard-godaddy, line 120: Stick to no inline code in headings per our [markdown guidelines](https://github.com/godaddy/gasket/blob/v7/CONTRIBUTING.md#markdown-documentation)
```suggestion
### middleware
```

packages/gasket-plugin-middleware/lib/index.d.ts
```diff
@@ -1,15 +1,73 @@
...
+    _implicitHeader: () => void;
+  }
+}
+
+declare module 'express' {
+  interface Request {
+    logger: {
+      metadata: Function;
+      child: Function;
+    };
```
agerard-godaddy, line 65: Could we just import and set the type from `@gasket/plugin-logger`?

```suggestion
    logger: Logger
```

packages/gasket-plugin-middleware/lib/index.d.ts
```diff
@@ -1,15 +1,73 @@
...
+     * compression middleware which is not present on the http2 request object.
+     * @see: https://github.com/expressjs/compression/pull/128
+     * and also, by the 'compiled' version in Next.js
+     * @see: https://github.com/vercel/next.js/issues/11669
+     */
+    _implicitHeader: () => void;
+  }
+}
+
+declare module 'express' {
```
agerard-godaddy, line 60: Do we also need to extend the Request of Fastify?

packages/gasket-plugin-middleware/lib/index.d.ts
```diff
@@ -1,15 +1,73 @@
...
+  interface Request {
+    logger: {
+      metadata: Function;
+      child: Function;
+    };
+  }
+}
 
-export = plugin;
```
agerard-godaddy, line 15: I believe we still need to export the [plugin shape](https://github.com/godaddy/gasket/blob/v7/packages/gasket-plugin-middleware/lib/index.d.ts#L10-L15).

packages/gasket-plugin-middleware/lib/index.d.ts
kbader-godaddy: Ah yeah that's fair. I suppose, to me, having all of the types in one file felt easier to maintain and organize. But I'm open to keeping them separate.

packages/gasket-plugin-middleware/lib/index.d.ts
kbader-godaddy: Ah yes! Great catch! Thanks!

packages/gasket-plugin-middleware/lib/index.d.ts
kbader-godaddy: Ah okay. Will update. I forget why exactly?

packages/gasket-plugin-middleware/lib/index.d.ts
kbader-godaddy: Well, actually, doesn't look like Logger has a `metadata` property.

packages/gasket-plugin-middleware/lib/index.d.ts
kbader-godaddy: Doesn't look like it.

agerard-godaddy (approved):

mmason2-godaddy (approved):

PR #881, opened 2024-09-04
Move actions in the logger plugin out of the hooks object

jpina1-godaddy:
## Summary

Move the gasket-plugin-logger actions from the hooks object to the gasket object

## Changelog

**@gasket/plugin-logger**
- Move actions out of the hooks object

## Test Plan

Update unit test

### DIFF
--- a/packages/gasket-plugin-logger/lib/index.js
+++ b/packages/gasket-plugin-logger/lib/index.js
@@ -24,6 +24,9 @@ module.exports = {
   name,
   version,
   description,
+  actions: {
+    getLogger: gasket => gasket.logger
+  },
   hooks: {
     create(gasket, { pkg, gasketConfig }) {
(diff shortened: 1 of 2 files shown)


packages/gasket-plugin-logger/lib/index.js
```diff
@@ -24,6 +24,9 @@ module.exports = {
   name,
   version,
   description,
+  actions: {
```
agerard-godaddy, line 27: Good catch

agerard-godaddy (approved):

packages/gasket-plugin-logger/lib/index.js
mmason2-godaddy: +1

mmason2-godaddy (approved):

PR #892, opened 2024-09-12
Handle route modules with default exports

agerard-godaddy:
## Summary

Handle route modules with default exports

## Changelog

**@gasket/plugin-express**
- Handle route modules with default exports

## Test Plan

Tested in local app


packages/gasket-plugin-express/lib/create-servers.js
kbader-godaddy, line 1: Worth adding a test to cover this?

jpage-godaddy (approved):

packages/gasket-plugin-express/lib/create-servers.js
agerard-godaddy: 👍 I pushed a fixtured with a default export.

Tests are currently failing with unrelated type issues - looking..

PR #905, opened 2024-09-17
V7 pr/chaos 1706@export prompt typings

mkasper1-godaddy:
## Summary

https://godaddy-corp.atlassian.net/browse/CHAOS-1706

having the create-gasket-app package export typings used on the prompts typings so the consuming apps can use the typings directly on new prompt hooks. This will allow the consuming app to resolve any issue related to typescript by casting needed args directly

## Changelog

- Created and exported typing `PromptUtils` so consuming apps could force proper typings on Args when creating prompts. This allows the consuming app to fix any odd typing issues that might popup

- Created and exported typing `PostCreatetUtils` so consuming apps could force proper typings on Args when creating prompts. This allows the consuming app to fix any odd typing issues that might popup


mkasper1-godaddy:
This work is not needed after @mmason2-godaddy update

PR #961, opened 2024-11-12
Adjustments to support Gasket in Next Edge Runtime

agerard-godaddy:
## Summary

There were some incompatibilities with Next's Edge Runtime ([docs](https://nextjs.org/docs/pages/api-reference/edge#unsupported-apis)) addressed by this PR. By having a lean export of config utils, we can avoid importing in Node API-based utils. `lodash.defaultsdeep` uses `new Function` which is incompatible, so I replaced with `deepmerge` which we use elsewhere and is [generally more widely used](https://npmtrends.com/deepmerge-vs-lodash.defaultsdeep).

## Changelog

**@gasket/core**
- Edge runtime support using lean export from `@gasket/utils`

**@gasket/utils**
- Adjustments to support Gasket in edge runtime

## Test Plan

Tested with App Router using Edge middleware


mmason2-godaddy (approved):

packages/gasket-core/lib/gasket.js
```diff
@@ -58,11 +58,11 @@ export class Gasket {
     // Can be used as a key to identify a gasket instance
     this.symbol = Symbol('gasket');
 
-    // @ts-ignore
+    // @ts-ignore - attached lifecycle trace methods not recognized
```
jpage-godaddy, line 61: You could add an `implements GasketEngine` or something like that on the class if you have things like `exec` in an interface.

agerard-godaddy: So, we already "extend" `GasketEngine` [here](https://github.com/godaddy/gasket/blob/edge-runtime/packages/gasket-core/lib/index.d.ts#L113). It works in consuming code.

agerard-godaddy: We took a look but didn't find a working option. I created a ticket to cycle back for an audit of these ts-ignores.

PR #963, opened 2024-11-13
fix(plugin-metadata): support esm only plugins and monorepo

aryasaatvik:

### DIFF
--- a/packages/gasket-plugin-metadata/lib/actions.js
+++ b/packages/gasket-plugin-metadata/lib/actions.js
@@ -49,6 +49,7 @@ async function getMetadata(gasket) {
     const presets = [];
     const modules = {};
 
+    // eslint-disable-next-line max-statements
     await gasket.execApply('metadata', async (plugin, handler) => {
       const isPreset = isGasketPreset.test(plugin.name);
       const isPlugin = isGasketPlugin.test(plugin.name);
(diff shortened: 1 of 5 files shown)


packages/gasket-plugin-metadata/lib/actions.js
```diff
@@ -62,7 +63,20 @@ async function getMetadata(gasket) {
         pluginData.metadata.path = path.join(app.metadata.path, 'plugins');
         plugins.push(pluginData);
       } else {
-        pluginData.metadata.path = path.dirname(path.join(require.resolve(pluginData.name), '..'));
+        let resolvedPath;
+        try {
+          resolvedPath = require.resolve(pluginData.name, { paths: [gasket.config.root] });
```
aryasaatvik, line 68: `{ paths: [gasket.config.root] }` allow any unhoisted dependencies to also get resolved. in a monorepo if plugin-metadata is a common it dependency and hoisted at the root and another dependency which uses the metadata hook but is only used by a single package in the monorepo this was throwing an error. starting the resolution at gasket app root fixes this

packages/gasket-plugin-metadata/lib/actions.js
```diff
@@ -62,7 +63,20 @@ async function getMetadata(gasket) {
...
+        } catch (err) {
+          // try esm import
+          try {
+            resolvedPath = await import(pluginData.name);
```
aryasaatvik, line 72: allows ESM only plugins to provide metadata

packages/gasket-plugin-metadata/lib/actions.js
```diff
@@ -62,7 +63,20 @@ async function getMetadata(gasket) {
...
+          // eslint-disable-next-line no-catch-shadow, no-shadow
+          } catch (err) {
+            gasket.logger.error(`Error resolving plugin ${pluginData.name}: ${err.message}`);
```
aryasaatvik, line 75: IMO errors resolving plugins for getting metadata should not crash the app but can revert if needed

packages/gasket-plugin-metadata/lib/actions.js
aryasaatvik: removed it lol

agerard-godaddy (approved):

mmason2-godaddy (approved):

PR #973, opened 2024-11-18
Normalized gasket request

agerard-godaddy:
## Summary

The purpose of this package is to provide a consistent request object for Gasket plugins and apps to use, regardless of the request handling framework.

## Changelog

**@gasket/request**
- Initial implementation

## Test Plan

N/A


kbader-godaddy (approved):

packages/gasket-request/LICENSE.md
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2019 GoDaddy Operating Company, LLC.
```
jpage-godaddy, line 3:
```suggestion
Copyright (c) 2024 GoDaddy Operating Company, LLC.
```

packages/gasket-request/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+The `GasketRequest` class is a representation of a request object that can be
+used by Gasket plugins and apps. It is a consistent shape that can be used
+across different request handling frameworks.
+
+A `GasketRequest` object has the following properties:
+- `headers` - (object) Request headers
+- [`cookies`] - (object) Request cookies
+- [`query`] - (object) Query parameters
+- [`path`] - (string) Request path
```
jpage-godaddy, line 26: Are these square brackets for showing these are optional? If so, when would `path` be missing? I also wonder wonder if it'd be easier if none of these were optional (path would default to `/`, cookies to `{}`, query to `{}`, etc).

packages/gasket-request/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+A `GasketRequest` can be created from a Node `IncomingMessage` object, an Express
+`Request` object, or a Next.js `NextRequest` object, amongst others.
+
+```js
+import { GasketRequest } from '@gasket/request';
+
+export default async function handler(req, res) {
+  const gasketRequest = new GasketRequest.make(req);
```
jpage-godaddy, line 35: Are these typos? I can see `new GasketRequest(...)` or `GasketRequest.make(...)` but not both in conjunction?

packages/gasket-request/README.md
```diff
@@ -0,0 +1,130 @@
...
+const headers = {
+  'x-example': 'example'
+};
+
+const staticGasketRequest = new GasketRequest.make({ headers });
+````
+
+### makeGasketRequest
+
+Another way to create a `GasketRequest` object is to use the `makeGasketRequest`.
```
jpage-godaddy, line 56: Do we need two ways to create these? Maybe we should just have one blessed gasket creation function in the module interface.

packages/gasket-request/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+export default async function handler(req, res) {
+  const gasketRequest = makeGasketRequest(req);
+  // use gasketRequest
+}
+```
+
+### withGasketRequest
+
+A higher-order function that wraps an action handler function and provides a
```
jpage-godaddy, line 69: This is for a Next.js server action, right? I'd clarify what you mean by "action" here. Also, is the name `withGasketRequest` specific enough to clarify that this is for server actions? Maybe this could live in `@gasket/next` and have a name like `gasketServerAction` or something like that.

jpage-godaddy: Hm, maybe this isn't about next.js specifically?

packages/gasket-request/README.md
agerard-godaddy: Yea, good callout. Reminds me I need to add some tests for path, also. It's possible for path NOT to exist for "static" GasketRequest objects for some prerendered page scenarios. Certainly not common. Perhaps we could default to empty string for path, with empty objects for cookies and query.

packages/gasket-request/README.md
agerard-godaddy: That's a typo

packages/gasket-request/lib/internal.d.ts
```diff
@@ -0,0 +1,8 @@
+import { Gasket } from '@gasket/core';
+import { RequestLike } from './index.js';
+
+export type withGasketRequest = (actionFn: (gasket: Gasket, req: RequestLike, ...args?: []) => Promise<unknown>) =>
+  (gasket: Gasket, req: RequestLike, ...args?: []) => Promise<unknown>
```
jpage-godaddy, line 5: I'd make sure you're preserving the types of the original function for the wrapped function; as this is currently it'll return a function that resolves to `Promise<unknown>` and has unknown parameter types. Also, shouldn't the inner function receive the normalized `GasketRequest` type?

```suggestion
export function withGasketRequest<Result, Args extends Array<unknown>>(actionFn: (gasket: Gasket, req: GasketRequest, ...args: Args) => Promise<Result>):
  (gasket: Gasket, req: RequestLike, ...args: Args) => Promise<Result>;
```

packages/gasket-request/README.md
agerard-godaddy: Yea, we don't need 2 ways. Probably wisest to only have one for discussion and debugging.

packages/gasket-request/README.md
agerard-godaddy: Its actually for GasketActions. I lifted this idea from your switchboard plugin. Will clarify in the docs

packages/gasket-request/lib/internal.d.ts
```diff
@@ -0,0 +1,8 @@
...
+
+export type withGasketRequestCache = (actionFn: (gasket: Gasket, req: RequestLike, ...args?: []) => Promise<unknown>) =>
+  (gasket: Gasket, req: RequestLike, ...args?: []) => Promise<unknown>
```
jpage-godaddy, line 8: Similar feedback.

```suggestion
export function withGasketRequestCache<Result, Args extends Array<unknown>>(actionFn: (gasket: Gasket, req: GasketRequest, ...args: Args) => Promise<Result>): (gasket: Gasket, req: RequestLike, ...args: Args) => Promise<Result>
```

packages/gasket-request/lib/request.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  query ??= {};
+
+  if (!normalizedReqMap.has(headers)) {
+    const normalized = new GasketRequest(Object.seal({
+      // @ts-ignore - entries not available on Headers here..
+      headers: headers.constructor.prototype.entries ? Object.fromEntries(headers.entries()) : headers,
+      cookies: cookies.constructor.prototype.getAll ? await objectFromCookieStore(cookies) : cookies,
+      query: query instanceof URLSearchParams ? Object.fromEntries(query) : query
```
jpage-godaddy, line 58: I wonder if you should document in the README that multiple values for a query parameter are not supported, only single-valued query parameters, since `URLSearchParams` does let you get an _array_ for a key.

packages/gasket-request/lib/request.js
```diff
@@ -0,0 +1,66 @@
...
+  let query = requestLike.query;
+
+  //
+  if (!query && 'nextUrl' in requestLike) {
+    query = requestLike.nextUrl.searchParams;
+  }
+
+  query ??= {};
+
+  if (!normalizedReqMap.has(headers)) {
```
jpage-godaddy, line 53: Not sure if this is worth addressing since it's mainly an optimization thing and duplication will probably lead to consistent results, but if you have two parallel requests awaiting a `makeGasketRequest` call they will duplicate efforts because we're not caching a Promise here.

packages/gasket-request/lib/wrappers.js
```diff
@@ -0,0 +1,26 @@
+import { GasketRequest } from './request.js';
+
+/** @type {import('./internal.d.ts').withGasketRequestCache} */
```
jpage-godaddy, line 3: If you make my suggested change of correcting this so it's an export of a function, not a type, you can do this instead:

```suggestion
/** @type {typeof import('./internal.d.ts').withGasketRequestCache} */
```

packages/gasket-request/lib/wrappers.js
```diff
@@ -0,0 +1,26 @@
...
+  return async (gasket, req, ...args) => {
+    const normalizedReq = await GasketRequest.make(req);
+
+    if (cache.has(normalizedReq)) {
+      return cache.get(normalizedReq);
+    }
+
+    const result = await actionFn(gasket, normalizedReq, ...args);
+    cache.set(normalizedReq, result);
+    return result;
```
jpage-godaddy, line 16: Similar feedback here; you're not going to be deduplicating parallel requests, only serial ones, unless you also cache the Promise. This should fix that issue.

```suggestion
    if (cache.has(normalizedReq)) {
      return cache.get(normalizedReq);
    }

    const promise = actionFn(gasket, normalizedReq, ...args);
    cache.set(normalizedReq, promise);
    
    try {
      const result = await promise;
      cache.set(normalizedReq, result);
    } catch (err) {
      cache.delete(normalizedReq);
      throw err;
    }

    return result;
```

packages/gasket-request/lib/wrappers.js
```diff
@@ -0,0 +1,26 @@
...
+  };
+}
+
+/** @type {import('./internal.d.ts').withGasketRequest} */
+export function withGasketRequest(actionFn) {
+  return async (gasket, req, ...args) => {
+    const normalizedReq = await GasketRequest.make(req);
+    return await actionFn(gasket, normalizedReq, ...args);
```
jpage-godaddy, line 24: Slightly more efficient because it avoids an additional Promise wrapper:

```suggestion
    return actionFn(gasket, normalizedReq, ...args);
```

packages/gasket-request/lib/internal.d.ts
jpage-godaddy: [Playground link](https://www.typescriptlang.org/play/?#code/C4TwDgpgBA4ghgZwNYWFAvFA3gXwFCiSyIrABKEAjgK4QJqa4HjQU13AAyAlihtvjwATCAGMANnABO0AGbUAdqODcA9gqgB3bsAAW8ZKja16AHgoJq44ABooAQSkBzBFAgAPYBAVDXjqXAgpopICqqaCgB8kQAUcMpqCgBiCgBcUDFOJKjpBqR2MpS52eRUJrZQAHTV0i7pji4AlBiRUAAKUqoAttwIEOZ0VsCRjal4UBlZhsDF0wVU6cYcPCh21ZW1CPXOCM3orR3dvf0WQ5F4eKLq9FAARpK6-Np6eUZlHHEIIEqTJfOUdlkqlU6QU1C6twgUjst2k6XoUm4Cice1aWHGEygMmA1CkGlhUkqMkgcGAMSBqkaAG48DhqRdEN9RFB5EoVOo7nBdDFmujMVArgobhT+HBNHAdHcHjFcHZZVAACx2ADk9y5yvp+CAA) that shows this approach and how the implicit typing preserves the original function parameter/return value typings.

agerard-godaddy: How did I not know about this playground... thank you!

packages/gasket-request/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+| Property | Type   | Description       | Arguments     |
+|----------|--------|-------------------|---------------|
+| headers  | object | Request headers   | required      |
+| cookies  | object | Request cookies   | default: `{}` |
+| query    | object | Query parameters  | default: `{}` |
+| path     | string | Request path      | default: `''` |
+
+
+The only required property to create a `GasketRequest` is the `headers` which
```
jpage-godaddy, line 32: Probably don't need this note for end-users since creation always goes through `makeGasketRequest` in the public interface.

packages/gasket-request/lib/index.d.ts
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Get a normalized request object for GasketActions
+ */
+export async function makeGasketRequest(req: RequestLike): Promise<GasketRequest>;
+
+type RequestActionFn<Result, Args extends Array<unknown>> = (gasket: Gasket, req: RequestLike, ...args?: Args) => Promise<Result>;
+type RequestActionWrapperFn<Result, Args extends Array<unknown>> = (gasket: Gasket, req: GasketRequest, ...args?: Args) => Promise<Result>;
+
+export async function withGasketRequest<Result, Args extends Array<unknown>>(actionFn: RequestActionFn<Result, Args>): RequestActionWrapperFn<Result, Args>;
+export async function withGasketRequestCache<Result, Args extends Array<unknown>>(actionFn: RequestActionFn<Result, Args>): RequestActionWrapperFn<Result, Args>;
```
jpage-godaddy, line 45: These are not async functions because they don't return a Promise wrapper around the wrapped functions.

```suggestion
export function withGasketRequest<Result, Args extends Array<unknown>>(actionFn: RequestActionFn<Result, Args>): RequestActionWrapperFn<Result, Args>;
export function withGasketRequestCache<Result, Args extends Array<unknown>>(actionFn: RequestActionFn<Result, Args>): RequestActionWrapperFn<Result, Args>;
```

packages/gasket-request/lib/request.js
```diff
@@ -0,0 +1,70 @@
...
+    throw new Error('request argument must have headers');
+  }
+
+  if (cache.has(headers)) {
+    return cache.get(headers);
+  }
+
+  const normalize = async () => {
+    let query = requestLike.query;
+    let path = requestLike.path;
```
jpage-godaddy, line 46: This property does not exist on [IncomingMessage](https://nodejs.org/docs/latest-v22.x/api/http.html#class-httpincomingmessage). There is a [`url` property](https://nodejs.org/docs/latest-v22.x/api/http.html#messageurl), however. Maybe we could have a parsed URL as part of the normalization to cover both path and query.

packages/gasket-request/lib/wrappers.js
```diff
@@ -0,0 +1,26 @@
...
+  return async (gasket, req, ...args) => {
+    const gasketRequest = await makeGasketRequest(req);
+
+    if (cache.has(gasketRequest)) {
+      return cache.get(gasketRequest);
+    }
+
+    const promise = actionFn(gasket, gasketRequest, ...args);
+    cache.set(gasketRequest, promise);
+    return await promise;
```
jpage-godaddy, line 16: Same feedback here about keeping Promises in cache. In my experience this can cause more memory retention issues, and also this doesn't take care of failed Promises being cached. Granted for the memory retention issue it's not as big of a deal since requests are short lived, but this should at least slim down the memory consumption when processing large amounts of requests concurrently.

```suggestion
    try {
      const result = await promise;
      cache.set(gasketRequest, result);
      return result;
    } catch (err) {
      cache.delete(gasketRequest);
      throw err;
    }
```

packages/gasket-nextjs/lib/request/index.js
```diff
@@ -0,0 +1,30 @@
...
+    return makeGasketRequest({
+      headers: headerStore,
+      cookies: cookieStore,
+      query
+    });
+  }
+
+  const promise = make();
+  reqCache.set(headerStore, promise);
+  return await promise;
```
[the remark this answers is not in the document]
agerard-godaddy: I created a `WeakPromiseKeeper` class which extends `WeakMap` to handle this logic in a DRY, testable way. Let me know if that covers the edge cases enough.

packages/gasket-request/README.md
```diff
@@ -0,0 +1,130 @@
...
+  async function handler(gasket, gasketRequest) {
+    // use gasketRequest
+  }
+);
+
+// example usage
+await gasket.actions.myAction(req);
+```
+
+If you action need additional arguments, you can pass them following the `gasketReq`.
```
jpina1-godaddy, line 77:
```suggestion
If your action needs additional arguments, you can pass them following the `gasketReq`.
```

packages/gasket-request/test/keeper.test.js
```diff
@@ -0,0 +1,67 @@
+import { expect } from '@jest/globals';
+import { WeakPromiseKeeper } from '../lib/keeper.js';
+
+const pause = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
```
jpage-godaddy, line 4: There's a node.js function for this already.

```
import { setTimeout } from 'node:timers/promises';
```

packages/gasket-nextjs/lib/request/index.js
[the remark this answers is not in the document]
jpage-godaddy: Yeah, I think this should work.

packages/gasket-request/lib/index.d.ts
```diff
@@ -0,0 +1,54 @@
...
+export async function makeGasketRequest(req: RequestLike): Promise<GasketRequest>;
+
+type RequestActionFn<Result, Args extends Array<unknown>> = (gasket: Gasket, req: RequestLike, ...args: Args) => Promise<Result>;
+type RequestActionWrapperFn<Result, Args extends Array<unknown>> = (gasket: Gasket, req: GasketRequest, ...args: Args) => Promise<Result>;
+
+export function withGasketRequest<Result, Args extends Array<unknown>>(actionFn: RequestActionFn<Result, Args>): RequestActionWrapperFn<Result, Args>;
+export function withGasketRequestCache<Result, Args extends Array<unknown>>(actionFn: RequestActionFn<Result, Args>): RequestActionWrapperFn<Result, Args>;
+
+
+const map = new WeakPromiseKeeper<GasketRequest, any>();
```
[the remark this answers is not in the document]
agerard-godaddy: 😳

jpage-godaddy (approved):

PR #974, opened 2024-11-19
add makeStoreOptionsFn

mkasper1-godaddy:
Updated the makeStoreOptions to be a callback that is callback or an object. This allows the makeStoreOptions to be triggered per request.

### DIFF
--- a/packages/gasket-redux/src/configure-make-store.js
+++ b/packages/gasket-redux/src/configure-make-store.js
@@ -34,25 +34,25 @@ export function prepareReducer(allReducers, rootReducer) {
  * @type {import('./index').configureMakeStore}
  */
 export default function configureMakeStore(makeStoreOptions = {}, postCreate) {
-  const {
-    thunkMiddleware = thunk,
-    middleware = [],
-    logging = false,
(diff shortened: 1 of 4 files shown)


packages/gasket-redux/src/configure-make-store.js
```diff
@@ -34,25 +34,25 @@ export function prepareReducer(allReducers, rootReducer) {
...
     const { req, logger = new Log() } = options;
+    const {
+      thunkMiddleware = thunk,
+      middleware = [],
+      logging = false,
+      enhancers = [(f) => f],
+      reducers = {},
+      rootReducer,
+      initialState = {}
+    } = typeof makeStoreOptions === 'function' ? makeStoreOptions(req) : makeStoreOptions;
```
agerard-godaddy, line 53: Would it be valuable to pass all the `makeStore` options through?
```suggestion
    } = typeof makeStoreOptions === 'function' ? makeStoreOptions(options) : makeStoreOptions;
```

agerard-godaddy (approved):

rmarkins-godaddy (approved):

PR #978, opened 2024-11-20
[PFX-813] ESM: Port plugin-metadata

jpina1-godaddy:
## Summary

Plugin metadata loads details from packages, which may be ESM. Since the plugin is written in CJS, this causes issues. Let’s first port this plugin to ESM so that we import both CJS and ESM package details with it.

## Changelog

**@gasket/plugin-metadata**
- Convert to ESM
- Convert jest tests to vitest

**gasket-typescript-tests**
- Update tests

## Test Plan

- Unit tests
- Tested locally


packages/gasket-plugin-metadata/lib/actions.js
```diff
@@ -1,13 +1,16 @@
...
 const isGasketModule = /(@gasket\/|gasket-)(?!plugin)(?!preset).+/;
 const isGasketPreset = /(gasket-preset)|(@gasket\/preset-)/;
 const isGasketPlugin = /(gasket-plugin)|(@gasket\/plugin-)/;
 let _metadata;
 
-function tryRequire(path) {
+function tryRequire(tryPath) {
   try {
-    return require(path);
+    return require(tryPath);
```
agerard-godaddy, line 13: The main reason this plugin should be is the first ESM port candidate is to handle loading metadata from other ESM plugins. Having a `tryRequire` for the `package.json` files _should_ be fine though. Is that the only use case `require()` besides `require.resolve()`?

jpina1-godaddy: Yes, require is using only for package.json and resolve. I saw this pattern in other plugins to handle package.json imports.

mmason2-godaddy: Yeah it's to get around the import attributes. If this works I'm good with it 👍 . I've never tried `require.resolve` with `createRequire` before but I imagine it would work.

mmason2-godaddy (approved):
I'd test this with the docs command and various app types and then 👍

packages/gasket-plugin-metadata/lib/actions.js
agerard-godaddy: Cool - I linked it into a working app and confirmed it works with the `docs` command:

Screenshot 2024-11-21 at 6 08 14 PM

agerard-godaddy (approved):
Nice work!

PR #982, opened 2024-11-22
Separate https-proxy plugin

agerard-godaddy:
## Summary

We have more use cases for an https proxy besides local development with Next.js built-in server. In particular, having an https proxy sidecar to deploy with Next.js apps for on-container HTTPS.

This introduces a new plugin, action and lifecycle with the https proxy features for a better separation of concerns for these scenarios.

## Changelog

**@gasket/plugin-https-proxy**
- Initial implementation

## Test Plan

Tested locally in app.

### DIFF
--- a/packages/gasket-plugin-https-proxy/README.md
+++ b/packages/gasket-plugin-https-proxy/README.md
@@ -1,0 +1,101 @@
+# @gasket/plugin-https-proxy
+
+Create an HTTPS proxy server with Gasket to use as a sidecar for Next.js apps 
+which do not handle HTTPS. This can be useful for local development when you 
+want to test HTTPS features and use secure cookies for authentication, etc.
+It can also be used in production if it is necessary for your application to
+handle HTTPS requests on the container.
+
+## Installation
+
+```
+npm i @gasket/plugin-https-proxy
+```
+
+Update your `gasket` file plugin configuration:
+
+```diff
+// gasket.js
+
++ import pluginHttpsProxy from '@gasket/plugin-https-proxy';
+
+export default makeGasket({
+  plugins: [
++   pluginHttpsProxy
+  ]
+});
+```
+
+## Configuration
+
+To be set under `httpsProxy` in the `gasket.js` file.
+
+This uses the [http-proxy] package and its full set of available [options].
+
+the `target` is 
+required to
+
+```diff
+// gasket.js
+export default makeGasket({
++  httpsProxy: {
++    protocol: 'https',
++    hostname: 'my-host.com',
++    port: 443,
++    xfwd: true,
++    ws: true,
++    target: {
++      host: 'localhost',
++      port: 80
++    }
++  }
+});
+```
+
+The above example forwards https requests on port `443` to `localhost:80`.
+
+The `protocol` and `hostname` are only used for logging about the proxy server
+and should not be confused with `target.protocol` and `target.host` which 
+are used for the actual destination server.
+
+## Actions
+
+### startProxyServer
+
+Use this action to start the HTTPS proxy server.
+
+```js
+import gasket from './gasket.js';
+gasket.actions.startProxyServer();
+```
+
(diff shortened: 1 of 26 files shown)


mmason2-godaddy (approved):

packages/gasket-plugin-https-proxy/README.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+## Configuration
+
+To be set under `httpsProxy` in the `gasket.js` file.
+
+This uses the [http-proxy] package and its full set of available [options].
+
+the `target` is 
+required to
```
jpage-godaddy, line 36: You accidentally left off

packages/gasket-plugin-https-proxy/README.md
```diff
@@ -0,0 +1,101 @@
...
++      host: 'localhost',
++      port: 80
++    }
++  }
+});
+```
+
+The above example forwards https requests on port `443` to `localhost:80`.
+
+The `protocol` and `hostname` are only used for logging about the proxy server
```
jpage-godaddy, line 57: I wonder if nesting things under a `source` property to complement `target` would make that more intuitive.

jpage-godaddy: Also, one thing that's a bit strange is that this is called an https proxy specifically but you have to specify the protocol as `https` as well. Should this be a general purpose proxy plugin instead?

packages/gasket-plugin-https-proxy/lib/actions.js
```diff
@@ -0,0 +1,26 @@
...
+  ).on('error', (err) => {
+    logger.error('Request failed to proxy:', err);
+  }).listen(
+    port
+  );
+
+  logger.info(`Proxy server started: ${protocol}://${hostname}:${port}`);
+}
+
+export default {
```
jpage-godaddy, line 24: Do we need a default export here? The importer can do an `import *` if they want to get all the exports.

packages/gasket-plugin-https-proxy/lib/index.d.ts
```diff
@@ -0,0 +1,33 @@
...
+  export interface GasketConfig {
+    httpsProxy?: HttpsProxyConfig
+  }
+
+  export interface HookExecTypes {
+    httpsProxy(proxyConfig: HttpsProxyConfig): MaybeAsync<HttpsProxyConfig>,
+  }
+
+  export interface GasketActions {
+    startProxyServer: () => Promise<void>;
```
jpage-godaddy, line 24: There may be a use case for the caller to receive the proxy server object and have the action return that instead of this being a `Promise<void>`.

packages/gasket-plugin-https-proxy/test/actions.test.js
```diff
@@ -0,0 +1,81 @@
+import { vi, describe, it, expect, beforeEach, afterEach } from 'vitest';
```
jpage-godaddy, line 1: Sweet. I'm also a big fan of `vitest`; it finally made modules just work.

jpage-godaddy (approved):
LGTM, just some minor suggestions.

packages/gasket-plugin-https-proxy/README.md
agerard-godaddy: Yes it is mostly pass-through configuration. I could default the protocol to `https` - I'll make that change. So, technically yea, the underlying `http-proxy` package could be used to proxy to other services or something, but I don't see it being useful that way in a Gasket context. I went back and forth on the name, but an HTTPS proxy is the specific gap intended to be filled with this plugin.

packages/gasket-plugin-https-proxy/test/actions.test.js
agerard-godaddy: 👍 becoming a fan also

PR #985, opened 2024-12-05
[PFX-773]: Compile warning with App Router

bbetts-godaddy:
## Summary

Ticket: [PFX-773](https://godaddy-corp.atlassian.net/browse/PFX-773)

There is a warning when running `npm run build` in new app router applications. When compiling the application, the webpack plugin logs `Critical dependency: the request of a dependency is an expression`. Webpack looks for strings as the arguments to require and import statements. 

## Changelog

- change require statement in webpack plugin to use string interpolation

## Test Plan

- all tests pass


agerard-godaddy (approved):
Confirmed in a local app

jpina1-godaddy (approved):

PR #989, opened 2024-12-11
Command plugin fixup

mmason2-godaddy:
## Summary
- Update `prepare` hook to be `execWaterfall` and pass the `gasket.config`. This allows for async configure option for certain use cases.
- Opt of a sync `configure` hook in the command plugin. Execution of the `commands` hook is also sync.
- Update the `gasket.command` property to be under `gasket.config.command`
- Move the `parse` call of CommanderJS to the async `ready` hook in the command plugin. This allows for async config with `prepare` to be captured and applied.
- Clean old references to `gasket.command.id`
- Tune `applyConfigOverrrides` to persist the `commands` config values if `commandId` is not passed. This function runs multiple times in different packages and the config values were always trimmed on the first pass.

## Changelog
- Update the command plugin execution and tuning for the `configure` hook and `ready` hook invocation of the CLI.

## Test Plan
Tests updated


packages/gasket-plugin-command/lib/ready.js
```diff
@@ -1,18 +1,15 @@
...
-  }
+  });
+}
+
+export default {
+  timing: {
+    first: true
+  },
+  handler: ready
 }
```
agerard-godaddy, line 15: I don't think it is necessary to set the timing of this hook
```suggestion
export default ready;
```

packages/gasket-plugin-data/lib/index.d.ts
```diff
@@ -17,7 +16,8 @@ declare module '@gasket/core' {
   import { RequestLike, GasketRequest } from '@gasket/request';
 
   export interface GasketConfig {
-    data?: GasketDataDefinition
+    data?: GasketDataDefinition;
+    command?: string;
```
agerard-godaddy, line 20: Shouldn't `command` should be declared on GasketConfig in the commands plugin, not here in data plugin.

jpina1-godaddy (approved):

packages/gasket-plugin-command/lib/configure.js
```diff
@@ -0,0 +1,37 @@
+// @ts-nocheck
+/* eslint-disable no-unused-vars, no-sync */
+import { applyConfigOverrides } from '@gasket/utils';
+import { gasketBin, processCommand } from './cli.js';
+const isGasketCommand = /\/gasket\.(js|ts|cjs|mjs)$/;
```
agerard-godaddy, line 5: Thoughts on relaxing this regex a bit for smaller "gasket" situations like `tsx gasket-docs.ts docs`, or `node gasket.some-other-name-part.js my-cmd`?

```suggestion
const isGasketCommand = /gasket[.-\w]*\.(js|ts|cjs|mjs)$/;
```

agerard-godaddy (approved):

PR #991, opened 2024-12-12
Cli fix

mmason2-godaddy:
## Summary
Fix issue where the Commander instance was parsing the CLI args more than once during create.

## Changelog
- Add logic to the `ready` hook of the command plugin to only parse when the `gasket.config.command` is present
- Fixup dynamic-plugins plugin `prepare` hook to return the config

## Test Plan
Updated tests

### DIFF
--- a/packages/gasket-plugin-dynamic-plugins/lib/index.js
+++ b/packages/gasket-plugin-dynamic-plugins/lib/index.js
@@ -10,23 +10,23 @@ export default {
   version,
   hooks: {
     async prepare(gasket, config) {
-      if (!gasket.config.dynamicPlugins?.length) return config;
+      if (!config.dynamicPlugins?.length) return config;
 
-      const load = await Promise.all(gasket.config.dynamicPlugins.map(plugin => import(plugin)));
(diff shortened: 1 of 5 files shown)


packages/gasket-plugin-dynamic-plugins/lib/index.js
```diff
@@ -9,8 +9,8 @@ export default {
   name,
   version,
   hooks: {
-    async prepare(gasket) {
-      if (!gasket.config.dynamicPlugins?.length) return;
+    async prepare(gasket, config) {
+      if (!gasket.config.dynamicPlugins?.length) return config;
```
agerard-godaddy, line 13: Since we are making `prepare` a waterfall lifecycle, you should probably modify and return the incoming `config` object here instead of mutating `gasket.config` as it will end up getting stomped over.

```suggestion
      if (!config.dynamicPlugins?.length) return config;
```

Basically, replace `gasket.config` -> `config` throughout this hook

jpina1-godaddy (approved):

agerard-godaddy (approved):

PR #1002, opened 2024-12-19
fix: Added isMergeableObject config to deepmerge

kbader-godaddy:
## Summary

The `deepmerge` function was only copying object properties instead of the whole object. We had an instance where methods on the winston format config were getting lost in the merge.

## Changelog

Added isMergeableObject config to deepmerge

## Test Plan

Existing tests all pass.

### DIFF
--- a/packages/gasket-utils/lib/config.js
+++ b/packages/gasket-utils/lib/config.js
@@ -1,20 +1,20 @@
 const deepmerge = require('deepmerge');
 // @ts-ignore - diagnostics lib does not have a types declaration file
 const debug = require('diagnostics')('gasket:utils');
+// @ts-ignore - 'isPlainObject' is not imported by using a default import
+const { isPlainObject } = require('is-plain-object');
 
 /**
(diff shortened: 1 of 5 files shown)


jpina1-godaddy (approved):

packages/gasket-utils/lib/config.js
```diff
@@ -1,20 +1,20 @@
...
-  { env = '', commandId }
-) {
+function applyConfigOverrides(config, { env = '', commandId }) {
   // @ts-ignore - merged config definitions
   return deepmerge.all(
-  // @ts-ignore - partial config definitions
-    [...getPotentialConfigs(config, { env, commandId })].reverse()
+    // @ts-ignore - partial config definitions
+    [...getPotentialConfigs(config, { env, commandId })].reverse(),
+    { isMergeableObject: isPlainObject }
```
agerard-godaddy, line 17: Could you add a test verifying that non-plain objects are copied correctly?

PR #1026, opened 2025-01-31
Allow partial configs in environment overrides

jpage-godaddy:
## Summary

Fix the TypeScript types so that you don't have to do a full set of config keys in environment overrides to avoid TypeScript errors.

## Test Plan

Added type test

### DIFF
--- a/packages/create-gasket-app/lib/commands/create.js
+++ b/packages/create-gasket-app/lib/commands/create.js
@@ -118,7 +118,7 @@ createCommand.action = async function run(appname, options, command) {
 
     const pluginGasket = makeGasket({
       ...context.presetConfig,
-      plugins: context.presets.concat(context.presetConfig.plugins)
+      plugins: context.presetConfig.plugins.concat(context.presets)
     });
 
     await promptHooks({ gasket: pluginGasket, context });
--- a/packages/gasket-core/lib/gasket.js
+++ b/packages/gasket-core/lib/gasket.js
@@ -38,7 +38,7 @@ export class Gasket {
     // prune nullish and/or empty plugins
     config.plugins = config.plugins
       .filter(Boolean)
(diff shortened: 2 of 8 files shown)


packages/create-gasket-app/lib/commands/create.js
```diff
@@ -118,7 +118,7 @@ createCommand.action = async function run(appname, options, command) {
 
     const pluginGasket = makeGasket({
       ...context.presetConfig,
-      plugins: context.presets.concat(context.presetConfig.plugins)
+      plugins: context.presetConfig.plugins.concat(context.presets)
```
jpage-godaddy, line 121: This is just an annoying limitation of TypeScript. Because `context.presets` is of type `Array<Plugin>` it cannot concatenate in `Array<Plugin | { default: Plugin }>` because `concat` is defined such where the return value is supposed to be the same as the original array. However, it can do this in the reverse order. I assume that the order of plugins doesn't matter.

packages/gasket-core/lib/gasket.js
```diff
@@ -38,7 +38,7 @@ export class Gasket {
     // prune nullish and/or empty plugins
     config.plugins = config.plugins
       .filter(Boolean)
-      .map(plugin => plugin.default || plugin) // quality of life for cjs apps
+      .map(plugin => 'default' in plugin ? plugin.default : plugin) // quality of life for cjs apps
```
jpage-godaddy, line 41: TypeScript is able to use an `in` check to narrow a type; seeing if a `default` property is falsy doesn't do the same, so the old expression wasn't able to be resolved as a `Plugin` object.

mmason2-godaddy (approved):

agerard-godaddy (approved):

PR #1028, opened 2025-02-04
Default https|http2 root property, cleanup plugin

mmason2-godaddy:
## Summary
- [configure hook](https://github.com/godaddy/gasket/blob/https-plugin-tuneup/packages/gasket-plugin-https/lib/configure.js#L2) to see the `root` property by default if it's not set. This is a quality of life change to avoid errors when defining certs inside the `https|http2` config properties.
- Clean up plugin

## Changelog
- Add configure hook to set a default for the `root` property inside the `https|htt2` config options

## Test Plan
Tests updated


packages/gasket-plugin-https/lib/configure.js
```diff
@@ -0,0 +1,22 @@
...
+module.exports = function configure(gasket, config) {
+  const { root } = config;
+
+  function setRoot(serverConfig) {
+    if (!serverConfig) return;
+    if (serverConfig.root) return;
+
+    if (Array.isArray(serverConfig)) {
+      serverConfig.forEach(item => {
+        item.root = root;
```
agerard-godaddy, line 11: In the array, you should probably only conditionally set `root` since you haven't checked for it
```suggestion
        item.root ??= root;
```

agerard-godaddy (approved):

kbader-godaddy (approved):

PR #1037, opened 2025-02-25
Migrate Monorepo to PNPM and Changesets

kbader-godaddy:


package.json
```diff
@@ -4,27 +4,31 @@
...
-    "postinstall": "npm run build",
+    "postinstall": "pnpm run build",
     "postpublish": "node scripts/mark-deprecated.mjs",
-    "publish": "lerna publish",
-    "publish:canary": "lerna publish --preid canary --dist-tag canary",
-    "publish:next": "lerna publish prerelease --preid next --dist-tag next --force-publish --exact",
-    "test:package": "lerna run test --stream --scope $*",
-    "typecheck:all": "lerna run typecheck --stream"
+    "publish": "pnpm changeset publish",
+    "publish:canary": "pnpm changeset version --snapshot canary && pnpm changeset publish --tag canary",
```
agerard-godaddy, line 28: I had to look up [snapshot docs](https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md). Sounds like an interesting for canary versions.

packages/gasket-plugin-analyze/lib/index.d.ts
```diff
@@ -10,9 +10,6 @@ declare module '@gasket/core' {
...
 }
 
-const plugin: Plugin = {
-  name: '@gasket/plugin-analyze',
-  hooks: {}
-};
+declare const plugin: Plugin;
 
-export = plugin;
+export default plugin;
```
agerard-godaddy, line 15: You have several plugin type changing this to default exports. However, we've had type issues because these packages are not `type: module`. What's the reason for the change now and have you vetted the imports with a TypeScript app?

kbader-godaddy: I am seeing these errors Screenshot 2025-03-04 at 10 59 01 AM

agerard-godaddy: Ok - It indeed reads better. Well, if it works, it works.

agerard-godaddy (approved):
Let do it!

mmason2-godaddy (approved):

PR #1044, opened 2025-03-05
Use Chosen Package Manager (Catalog)

kbader-godaddy:
## Summary

When creating a new gasket app, gasket was forcing users to use pnpm instead of the package manager that was chosen by the user. 

Also, updated deps.

## Test Plan

 
Updated tests.


packages/create-gasket-app/lib/scaffold/actions/global-prompts.js
```diff
@@ -37,7 +37,8 @@ async function choosePackageManager(context, prompt) {
             { name: 'npm' },
             { name: 'pnpm' },
             { name: 'yarn' }
-          ]
+          ],
+          default: 'pnpm'
```
agerard-godaddy, line 41: I think defaulting to npm will be safest for end users

mmason2-godaddy (approved):

agerard-godaddy (approved):

PR #1061, opened 2025-03-21
Update to react 19 and next 15

jpina1-godaddy:
## Summary

Update to react 19 and next 15

## Test Plan

Tested locally


packages/gasket-plugin-nextjs/package.json
```diff
@@ -78,17 +78,17 @@
     "eslint-plugin-unicorn": "^55.0.0",
     "jest": "^29.7.0",
     "lodash.merge": "^4.6.2",
-    "next": "^14.2.24",
+    "next": "15.2.3",
```
agerard-godaddy, line 81: Can we use the range here?
```suggestion
    "next": "^15.2.3",
```

packages/gasket-plugin-redux/lib/index.d.ts
```diff
@@ -39,12 +39,6 @@ declare class ReduxReducers {
   addEntry(str: string): void;
 }
 
-declare module 'http' {
```
agerard-godaddy, line 42: Not that I disagree with removing this, but why the change in this PR?

site/package.json
```diff
@@ -20,12 +20,12 @@
...
-    "react": "catalog:",
-    "react-dom": "^18.0.0"
+    "react": "^19.0.0",
+    "react-dom": "^19.0.0"
   },
   "devDependencies": {
-    "@docusaurus/module-type-aliases": "3.1.1",
-    "@docusaurus/types": "3.1.1"
+    "@docusaurus/module-type-aliases": "3.7.0",
+    "@docusaurus/types": "3.7.0"
```
agerard-godaddy, line 28: Should be able to use the ^ range here now

packages/gasket-plugin-redux/lib/index.d.ts
jpina1-godaddy: Typescript was throwing an error because it was already defined in gasket-redux

packages/gasket-redux/test/configure-make-store.spec.js
```diff
@@ -1,37 +1,15 @@
-import * as redux from 'redux';
 import configureMakeStore, { prepareReducer } from '../src/configure-make-store';
-import thunk from 'redux-thunk';
-import { mockLoggerMiddleware } from 'redux-logger';
-
-const combineReducersSpy = jest.spyOn(redux, 'combineReducers');
-const applyMiddlewareSpy = jest.spyOn(redux, 'applyMiddleware');
-const createStoreSpy = jest.spyOn(redux, 'createStore');
-const composeSpy = jest.spyOn(redux, 'compose');
+import { withExtraArgument } from 'redux-thunk';
```
agerard-godaddy, line 2: You are importing `thunk` in the src file. These should match up I would think

jpina1-godaddy: Yeah it's for the test here https://github.com/godaddy/gasket/pull/1061/files/c23b4d6eca3c93423bc6680133936fb616ff790c#diff-d47cbaa94c5d46e16a5e07c13985d4f10230089226c172864bd4cabf451579e7R101. Thunk's exports changed.

agerard-godaddy (approved):

kbader-godaddy (approved):

PR #1085, opened 2025-04-17
Throw if hook returns promise for sync exec method

agerard-godaddy:
## Summary

Now and then, we will have an app with plugin hooks that are async when they shouldn't be. This tends to pop up most with the `configure` lifecycle. To help us catch and debug sooner, we will throw when promises are returned for sync methods.

## Test Plan

Updated unit tests


jpage-godaddy (approved):
Seems reasonable

PR #1098, opened 2025-05-05
[PFX-919] - force eslint-plugin-react-hooks version on different code styles

mmason2-godaddy:
## Summary
Strange issue where PNPM would install two versions of `eslint-plugin-react-hooks` regardless of peer deps. This was causing warnings in the create flow with the lint script. I've added `eslint-plugin-react-hooks` as a dev dep in the relevant code styles.

## Test Plan
Added test case


kbader-godaddy (approved):

.changeset/ninety-cows-work.md
```diff
@@ -0,0 +1,5 @@
+---
+"@gasket/plugin-lint": minor
```
agerard-godaddy, line 2: I think this should just be a patch fix
```suggestion
"@gasket/plugin-lint": patch
```

agerard-godaddy (approved):

PR #1120, opened 2025-05-21
Fix middleware logic

mmason2-godaddy:
## Summary
Middleware was busted after TS audit changes

## Test Plan
Passing

### DIFF
--- a/packages/gasket-plugin-middleware/lib/utils.js
+++ b/packages/gasket-plugin-middleware/lib/utils.js
@@ -1,27 +1,19 @@
 const cookieParser = require('cookie-parser');
 const compression = require('compression');
+// @ts-ignore
 const diagnostics = require('diagnostics');
 
 const debug = diagnostics('gasket:middleware');
-
-/**
- * Type guard to detect if the app is an Express app.
- * @type {import('./internal').canUseMiddleware}
- */
-function canUseMiddleware(app) {
-  return typeof app?.use === 'function' && typeof app?.set === 'function';
-}
 
 /**
  * Applies the cookie parser based on the middleware pattern.
  * @type {import('./internal').applyCookieParser}
  */
 function applyCookieParser(app, middlewarePattern) {
-  const middleware = cookieParser();
   if (middlewarePattern) {
-    app.use(middlewarePattern, middleware);
+    app.use(middlewarePattern, cookieParser());
   } else {
-    app.use(middleware);
+    app.use(cookieParser());
   }
 }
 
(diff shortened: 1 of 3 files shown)


packages/gasket-plugin-middleware/lib/utils.js
```diff
@@ -89,14 +82,8 @@ async function executeMiddlewareLifecycle(gasket, app, middlewarePattern) {
...
 
-  if (canUseMiddleware(app)) {
-    applyCookieParser(app, middlewarePattern);
-    applyCompression(app, config.compression);
-  } else {
-    debug('Skipping Express-specific middleware: app is not Express');
-  }
-
   await gasket.execApply('middleware', async (plugin, handler) => {
+    // @ts-ignore - TS is unable to infer that `app` can be either an Express application or a Fastify instance
```
jpage-godaddy, line 86: Why is that? Can't you make the type of `app` something like `ExpressApp | FastifyApp`?

packages/gasket-plugin-middleware/lib/utils.js
```diff
@@ -52,30 +44,31 @@ function applyMiddlewareConfig(middleware, plugin, middlewareConfig, middlewareP
   const configEntry = middlewareConfig.find(mw => mw.plugin === pluginName);
 
   if (configEntry) {
-    middleware.paths = configEntry.paths || [];
+    // @ts-ignore
```
jpage-godaddy, line 47: Why ignore these? If you want to build objects of this type, shouldn't you define that type?

packages/gasket-plugin-middleware/lib/utils.js
```diff
@@ -52,30 +44,31 @@ function applyMiddlewareConfig(middleware, plugin, middlewareConfig, middlewareP
...
 }
 
 /**
  * Attaches the middleware layers to the app.
- * @type {import('./internal').applyMiddlewaresToApp}
+ *  @type {import('./internal').applyMiddlewaresToApp}
  */
 function applyMiddlewaresToApp(app, middlewares, middlewarePattern) {
-  if (!canUseMiddleware(app)) {
-    debug('Skipping middleware application – not an Express app');
```
jpage-godaddy, line 68: Were these checks to protect Fastify apps? It's not clear to me if making these checks will not cause problems.

packages/gasket-plugin-middleware/lib/utils.js
mmason2-godaddy: I didn't dive into the TS types. I just restored it back to how it was when it wasn't busted. We can look at them at a later date unless you want to contribute to the PR.

packages/gasket-plugin-middleware/lib/utils.js
mmason2-godaddy: I didn't make the changes and I'm not familiar. I was out on leave when they occurred. @kbader-godaddy made the changes.

kbader-godaddy:
@jpage-godaddy https://github.com/godaddy/gasket/pull/1121

PR #1121, opened 2025-05-21
Support mixed middleware formats in app.use()

kbader-godaddy:
## Summary

Fixes a bug where object-style middleware entries (e.g. `{ handler, paths }`) or arrays of middleware functions were being passed directly to `app.use()`, which expects a function. This caused runtime errors during the middleware lifecycle in apps where plugins returned unsupported formats.

This update ensures that:

* Function-style, object-style, and array-wrapped middleware are all handled consistently
* Invalid entries (e.g. missing or non-function `handler`) are safely ignored
* Middleware config entries properly append `middlewarePattern` to `paths` when applicable

## Test Plan

* Added unit tests to verify support for:
  * Function-style middleware
  * Object-style middleware with `paths`
  * Arrays of middleware functions
  * Invalid middleware entries that normalize to `{ handler: undefined }`
* Verified that middleware config entries correctly combine configured paths with the `middlewarePattern`
* Confirmed `app.use()` is only called with valid, normalized handlers

### DIFF
--- a/packages/gasket-plugin-middleware/lib/internal.d.ts
+++ b/packages/gasket-plugin-middleware/lib/internal.d.ts
@@ -89,7 +89,7 @@ export function isMiddlewareObject(value: any): value is { handler: Handler; pat
  * @returns {{ handler: Handler | undefined, paths?: string | RegExp | Array<string | RegExp> }}
  */
 export function normalizeMiddlewareEntry(
-  entry: any
+  entry: unknown
 ): {
   handler: Handler | undefined;
   paths?: string | RegExp | Array<string | RegExp>;
--- a/packages/gasket-plugin-middleware/lib/internal.d.ts
+++ b/packages/gasket-plugin-middleware/lib/internal.d.ts
@@ -72,17 +72,78 @@ export function executeMiddlewareLifecycle(
  */
 export function attachLogEnhancer(req: ExpressRequest): void;
 export function isValidMiddleware(middleware: Function | Function[]): boolean;
+
+/**
+ * Checks whether a given value is a middleware object with a `handler` function.
+ *
+ * @param {any} value - The value to check
+ * @returns {value is { handler: Handler; paths?: string | RegExp | Array<string | RegExp> }} True if the value is a middleware object
+ */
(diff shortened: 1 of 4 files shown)


agerard-godaddy (approved):

mmason2-godaddy (approved):

jpage-godaddy (approved):

packages/gasket-plugin-middleware/lib/utils.js
```diff
@@ -69,7 +70,14 @@ function applyMiddlewaresToApp(app, middlewares, middlewarePattern) {
     return;
   }
 
-  middlewares.forEach(({ handler, paths }) => {
+  middlewares.forEach((layer) => {
+    // Normalize: allow either direct middleware function or object with { handler, paths }
+    const paths = layer.paths;
+    const handler = typeof layer === 'function' ? layer : layer.handler;
```
jpage-godaddy, line 76: Can't `layer` also be an Array of middleware functions?

jpage-godaddy (changes requested):

packages/gasket-plugin-middleware/lib/utils.js
agerard-godaddy: Yes it could, good catch.

kbader-godaddy: Great catch! Thanks @jpage-godaddy !

packages/gasket-plugin-middleware/lib/internal.d.ts
```diff
@@ -72,17 +72,78 @@ export function executeMiddlewareLifecycle(
...
+
+/**
+ * Normalizes a middleware entry into a consistent `{ handler, paths }` format.
+ * Accepts either a function, an object with a `handler`, or an invalid entry.
+ *
+ * @param {any} entry - The middleware entry to normalize
+ * @returns {{ handler: Handler | undefined, paths?: string | RegExp | Array<string | RegExp> }}
+ */
+export function normalizeMiddlewareEntry(
+  entry: any
```
jpage-godaddy, line 92: If we don't have a more specific type for `entry`, `unknown` is better than `any`.

```suggestion
  entry: unknown
```

jpage-godaddy (approved):
Looks more correct now, thanks.

PR #1133, opened 2025-06-05
Revive Preboot

mmason2-godaddy:
## Summary

Slack thread: https://godaddy.slack.com/archives/CABCTNQ5P/p1749076141557509

### What changed

- Added a `prepareProxyServer` action to `@gasket/plugin-https-proxy` and a `prepareServer` action to `@gasket/plugin-https`. Both ensure the `preboot` lifecycle runs before starting the (proxy) server.
- Updated type definitions to include the new actions.
- Updated plugin metadata to document the new actions and the `preboot` lifecycle.
- Expanded README docs for both plugins to document the new actions and lifecycle, including usage examples.
- Refactored `startProxyServer` and `startServer` to call their respective `prepare*` actions instead of duplicating preboot logic.
- Added/expanded tests for the new actions, including error propagation and correct lifecycle execution.

### Standalone Default Server

It's possible to wire up a similar circumstance when only using NextJS default server. A user could execute standalone script to run precursor operations before the `next start|dev` call. 

---

**Summary:**  
This commit ensures that a `preboot` lifecycle is always executed before starting either the main server or the HTTPS proxy server, adds/updates types and docs, and improves test coverage for these flows.

## Test Plan
Added and updated. Tested `preboot` locally in custom server and default server with a proxy.

### DIFF
--- a/.changeset/smart-parks-remain.md
+++ b/.changeset/smart-parks-remain.md
@@ -1,5 +1,0 @@
----
-"@gasket/request": patch
----
-
-Parse cookie header if cookies not on request-like object
--- a/packages/create-gasket-app/CHANGELOG.md
+++ b/packages/create-gasket-app/CHANGELOG.md
@@ -1,5 +1,19 @@
 # `create-gasket-app`
 
+## 7.4.4
+
+### Patch Changes
+
+- Updated dependencies [39e41cb]
+  - @gasket/request@7.4.1
+  - @gasket/plugin-command@7.4.0
+  - @gasket/plugin-docs@7.4.1
+  - @gasket/plugin-docusaurus@7.4.2
+  - @gasket/plugin-dynamic-plugins@7.3.7
+  - @gasket/plugin-git@7.4.0
+  - @gasket/plugin-logger@7.3.4
+  - @gasket/plugin-metadata@7.4.2
+
 ## 7.4.3
 
 ### Patch Changes
--- a/packages/create-gasket-app/package.json
+++ b/packages/create-gasket-app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "create-gasket-app",
-  "version": "7.4.3",
+  "version": "7.4.4",
   "description": "starter pack for creating a gasket app",
   "type": "module",
   "bin": {
--- a/packages/gasket-data/CHANGELOG.md
+++ b/packages/gasket-data/CHANGELOG.md
@@ -1,5 +1,12 @@
 # `@gasket/data`
 
+## 7.4.1
+
+### Patch Changes
+
+- Updated dependencies [39e41cb]
+  - @gasket/request@7.4.1
+
 ## 7.4.0
 
 ### Minor Changes
--- a/packages/gasket-data/package.json
+++ b/packages/gasket-data/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gasket/data",
-  "version": "7.4.0",
+  "version": "7.4.1",
   "description": "Helper package for accessing embedded Gasket Data in the browser",
   "type": "module",
   "types": "lib/index.d.ts",
--- a/packages/gasket-nextjs/CHANGELOG.md
+++ b/packages/gasket-nextjs/CHANGELOG.md
@@ -1,5 +1,13 @@
 # `@gasket/nextjs`
 
+## 7.5.1
+
+### Patch Changes
+
+- Updated dependencies [39e41cb]
+  - @gasket/request@7.4.1
+  - @gasket/data@7.4.1
+
 ## 7.5.0
 
 ### Minor Changes
--- a/packages/gasket-nextjs/package.json
+++ b/packages/gasket-nextjs/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gasket/nextjs",
-  "version": "7.5.0",
+  "version": "7.5.1",
   "description": "Gasket integrations for Next.js apps",
   "type": "module",
   "types": "lib/index.d.ts",
--- a/packages/gasket-plugin-data/CHANGELOG.md
+++ b/packages/gasket-plugin-data/CHANGELOG.md
@@ -1,5 +1,12 @@
 # `@gasket/plugin-data`
 
+## 7.4.1
+
+### Patch Changes
+
+- Updated dependencies [39e41cb]
+  - @gasket/request@7.4.1
+
 ## 7.4.0
 
 ### Minor Changes
--- a/packages/gasket-plugin-data/package.json
+++ b/packages/gasket-plugin-data/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gasket/plugin-data",
-  "version": "7.4.0",
+  "version": "7.4.1",
   "description": "Supports application-specific settings and configurations",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
--- a/packages/gasket-plugin-elastic-apm/CHANGELOG.md
+++ b/packages/gasket-plugin-elastic-apm/CHANGELOG.md
@@ -1,5 +1,12 @@
 # `@gasket/plugin-elastic-apm`
 
+## 7.4.1
+
+### Patch Changes
+
+- Updated dependencies [39e41cb]
+  - @gasket/request@7.4.1
+
 ## 7.4.0
 
 ### Minor Changes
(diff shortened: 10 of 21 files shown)


packages/gasket-plugin-https-proxy/README.md
```diff
@@ -104,6 +113,22 @@ gasket.actions.startProxyServer();
 
 ## Lifecycles
 
+### preboot
```
jpage-godaddy: I think something as essential as "things I want to do before my server starts" would be misplaced being part of `@gasket/plugin-https-proxy` which seems like a plugin that would not be necessarily a universal thing but more like a weird plugin we have to work around next.js specificities. I was thinking it should be part of `@gasket/plugin-https` instead (which would be better named `@gasket/plugin-web-server`).

jpage-godaddy: Actually, it looks like you did put the lifecycle in `@gasket/plugin-https`; maybe the docs should be moved there as well.

mmason2-godaddy: I've created both to allow for this usecase with custom server and default server with a proxy(which is the default for all webapp variations).

mmason2-godaddy: It would be ideal to have the `preboot` lifecycle and calling action to be one place. This presents issues because of two https plugins now. I could create a new plugin only for `preboot` but that seemed like overkill honestly. `plugin-https` is a dependency of `plugin-https-proxy` but only for type reasons. In default server proxy apps `plugin-https` is not actually added or instantiated in the Gasket at runtime.

kbader-godaddy:
@mmason2-godaddy just to confirm, the preboot lifecycle was never used in these 2 plugins, correct? It was a part of the `@gasket/plugin-elastic-apm` before the v7 launch?

packages/gasket-plugin-https-proxy/README.md
```diff
@@ -93,6 +93,15 @@ export default makeGasket({
...
 
+### prepareProxyServer
+
+This action allows for plugins to execute operations before the proxy server is started. It will execute the `preboot` lifecycle.
+
+```js
+import gasket from './gasket.js';
+gasket.actions.prepareProxyServer();
+```
+
```
kbader-godaddy: Would this be an appropriate comment?
```suggestion

> **Best Practice**: This action is automatically called by `startProxyServer`, so you typically don't need to call it directly unless you need to run preparation steps separately.
```

packages/gasket-plugin-https-proxy/README.md
```diff
@@ -104,6 +113,22 @@ gasket.actions.startProxyServer();
```
kbader-godaddy, line 104: Would this context be helpful?

```suggestion

This action follows the pattern: `prepareProxyServer` → `preboot` lifecycle → proxy server initialization.

The complete flow is:

1. Waits for `gasket.isReady` (ensuring async configuration is complete)
2. Executes the `preboot` lifecycle (one-time initialization)
3. Executes the `httpsProxy` lifecycle for dynamic configuration
4. Creates and starts the proxy server
```

packages/gasket-plugin-https-proxy/README.md
```diff
@@ -104,6 +113,22 @@ gasket.actions.startProxyServer();
...
+/**
+ * Executed before the proxy server is started.
+ *
+ * @param {Gasket} gasket Gasket API.
+ */
+preboot: async function preboot(gasket) {
+  // async operations
+}
+```
+
```
kbader-godaddy, line 129: Is this helpful or is it too obvious?

```suggestion

#### When to use `preboot` vs `prepare`

The `preboot` lifecycle runs **once** when the proxy server starts, while the `prepare` lifecycle runs for **every** Gasket instance creation. Use `preboot` for:

- One-time proxy server initialization
- Setting up shared SSL/TLS contexts
- Loading certificates that will be reused across requests
- Establishing persistent connections or pools

Use `prepare` for:

- Per-instance configuration
- Setting up instance-specific state
- Operations that need fresh initialization for each Gasket instance

> **Note**: In SSR applications, new Gasket instances may be created for different contexts, causing `prepare` to run multiple times. For proxy servers, this can lead to unnecessary overhead. Use `preboot` for expensive initialization operations.
```

packages/gasket-plugin-https/README.md
```diff
@@ -100,6 +100,15 @@ export default makeGasket({
...
 
+### prepareServer
+
+This action allows for plugins to execute operations before the server is started. It will execute the `preboot` lifecycle.
+
+```js
+import gasket from './gasket.js';
+gasket.actions.prepareServer();
+```
+
```
kbader-godaddy:
```suggestion

> **Best Practice**: This action is automatically called by `startServer`, so you typically don't need to call it directly unless you need to run preparation steps separately.
```
Same as above.

packages/gasket-plugin-https-proxy/README.md
mmason2-godaddy: This a great addition 🏅

packages/gasket-plugin-https-proxy/lib/actions.js
```diff
@@ -1,8 +1,19 @@
...
+ */
+export async function prepareProxyServer(gasket) {
+  await gasket.isReady;
+  await gasket.exec('preboot');
+}
+
 /** @type {import('@gasket/core').ActionHandler<'startProxyServer'>} */
 export async function startProxyServer(gasket) {
-  await gasket.isReady;
+  await gasket.actions.prepareProxyServer();
```
agerard-godaddy: I don't think we would ever want users execute the `preboot` outside the context of `startProxyServer`, would we? In which case the added `prepareProxyServer` action seem unnecessary. Instead, just kick off the preboot lifecycle when starting.

Also, having 2 different plugins capable of executing the same lifecycles seems dangerous. In theory, both the http and http-proxy plugins could be in used with the same app. Let's distinguish the lifecycles by name, which will also plugins differentiate what they're hooking as needed. Perhaps `preboot` for https plugin, and `prebootHttpsProxy` here?

```suggestion
  await gasket.isReady;
  await gasket.exec('prebootHttpsProxy');
```

packages/gasket-plugin-https/lib/actions.js
```diff
@@ -5,14 +5,25 @@ const one = require('one-time/async');
...
+
 /**
  * Gasket action: startServer
  * @param {import('@gasket/core').Gasket} gasket Gasket instance
  * @returns {Promise<void>} promise
  * @public
  */
 async function startServer(gasket) {
-  await gasket.isReady;
+  await gasket.actions.prepareServer();
```
agerard-godaddy: See other comment - removed the unnecessary action, execute `preboot` directly with a distinguishable name.

```suggestion
  await gasket.isReady;
  await gasket.exec('preboot');
```

agerard-godaddy (changes requested):
We shouldn't need the extra actions and distinguish between preboot hooks.

packages/gasket-plugin-https-proxy/README.md
```diff
@@ -102,8 +102,48 @@ import gasket from './gasket.js';
...
 ```
 
+This action follows the pattern: `prepareProxyServer` → `preboot` lifecycle → proxy server initialization.
+
+The complete flow is:
+
+1. Waits for `gasket.isReady` (ensuring async configuration is complete)
+2. Executes the `preboot` lifecycle (one-time initialization)
+3. Executes the `httpsProxy` lifecycle for dynamic configuration
+4. Creates and starts the proxy server
```
agerard-godaddy:
```suggestion

The complete flow is:

1. Waits for `gasket.isReady` (ensuring async configuration is complete)
2. Executes the `prebootHttpsProxy` lifecycle (one-time initialization)
3. Executes the `httpsProxy` lifecycle for dynamic configuration
4. Creates and starts the proxy server
```

packages/gasket-plugin-https-proxy/README.md
```diff
@@ -102,8 +102,48 @@ import gasket from './gasket.js';
...
+ * Executed before the proxy server is started.
+ *
+ * @param {Gasket} gasket Gasket API.
+ */
+prebootHttpsProxy: async function prebootHttpsProxy(gasket) {
+  // async operations
+}
+```
+
+#### When to use `preboot` vs `prepare`
```
agerard-godaddy: Adjust the lifecycle name in this doc block
```suggestion
#### When to use `prebootHttpsProxy` vs `prepare`
```

PR #1164, opened 2025-07-18
Type fixes to help with middleware

jpage-godaddy:
## Summary

- Make sure Express & Fastify handlers are considered compatible as middleware plugin handlers
- Make sure Express & Fastify requests are allowed for `RequestLike` values

This is to fix the fact that annoying casts are currently required when implementing a `middleware` hook and when calling `gasket.actions.checkAuth` when using Express types.

## Test Plan

Added TypeScript tests


jpage-godaddy:
Actually hold on... tried in my app and further fixes are required; will ping when ready.

mmason2-godaddy (approved):

agerard-godaddy (approved):

PR #1166, opened 2025-07-21
Eslint9

mmason2-godaddy:
# Upgrade to ESLint 9 with Flat Config Format

## 🎯 Overview
Upgrades the entire monorepo from ESLint 8 to ESLint 9, implementing the new flat configuration format and centralizing all ESLint configuration at the root level.

### Configuration Migration
- **New centralized config**: Created `eslint.config.js` at root using ESLint 9 flat config format
- **Removed individual configs**: Cleaned up `eslintConfig` sections from all 49 package.json files
- **Updated dependencies**: Centralized all ESLint-related dependencies in root `package.json`

### Dependency Updates
- `eslint`: `^8.57.1` → `^9.31.0` 
- `eslint-config-godaddy`: Updated to `^8.0.2` (ESLint 9 compatible)
- `eslint-config-godaddy-react`: Updated to `^10.0.0`
- `@typescript-eslint/eslint-plugin`: Updated to `^8.38.0`

### Configuration Features
- Uses modern flat config with spread syntax for config composition
- Maintains all existing GoDaddy ESLint configurations (base, React, TypeScript)
- Proper ignore patterns for build artifacts and test fixtures
- Jest and Vitest environment globals configured
- Unicorn and JSDoc plugins integrated

## 🧹 Cleanup
- Removed redundant ESLint dependencies from individual packages
- Eliminated duplicate configuration across packages
- Streamlined workspace ESLint management

## ✅ Verification
- [x] All packages lint successfully with `pnpm run lint:all`
- [x] No remaining `.eslintrc` files 
- [x] Configuration works across JavaScript, TypeScript, and React files
- [x] Maintains existing code style standards

## 🔧 Technical Notes
- Uses `defineConfig` from `eslint/config` for better TypeScript support
- Preserves all existing linting rules and patterns
- Compatible with existing CI/CD linting workflows
- No breaking changes to development experience


eslint.config.js
```diff
@@ -0,0 +1,50 @@
...
+import goddaddyReactTypescript from 'eslint-config-godaddy-react-typescript';
+import unicorn from 'eslint-plugin-unicorn';
+import vitest from '@vitest/eslint-plugin';
+import jsdoc from 'eslint-plugin-jsdoc';
+
+export default defineConfig([
+  ...godaddy,
+  ...goddaddyTypescript,
+  ...goddaddyReact,
+  ...goddaddyReactTypescript,
```
agerard-godaddy: It appears that the typescript configs also include the base godaddy ones. I can't tell if there's a cost for the redundancy, but in any case this works:

```suggestion
  ...goddaddyTypescript,
  ...goddaddyReactTypescript,
```

eslint.config.js
```diff
@@ -0,0 +1,50 @@
...
+      '**/node_modules/**',
+      '**/dist/**',
+      '**/cjs/**',
+      '**/react/**',
+      '**/generator/**',
+      '**/__mocks__/**',
+      '**/test/fixtures/**',
+      '**/gasket-redux/lib'
+    ]
+  },
```
agerard-godaddy: Having a single `ignores` property in a config makes it "global" - however, to clarify this, you could use the `globalIgnores` helper ([docs](https://eslint.org/docs/latest/use/configure/ignore#name-the-global-ignores-config)):

```suggestion
  globalIgnores([
      '**/node_modules/**',
      '**/dist/**',
      '**/cjs/**',
      '**/react/**',
      '**/generator/**',
      '**/__mocks__/**',
      '**/test/fixtures/**',
      '**/gasket-redux/lib'
    ]),
```

eslint.config.js
```diff
@@ -0,0 +1,50 @@
...
+    languageOptions: {
+      globals: {
+        ...vitest.environments.env.globals,
+        ...jest.environments.globals.globals,
+      },
+    },
+    rules: {
+      ...vitest.configs.recommended.rules,
+      'unicorn/filename-case': 'error',
+      'no-sync': 'off',
```
agerard-godaddy: Is this a new rule, or why was it disabled? Could we get away with `warn`?

eslint.config.js
```diff
@@ -0,0 +1,50 @@
...
+      jsdoc
+    },
+    languageOptions: {
+      globals: {
+        ...vitest.environments.env.globals,
+        ...jest.environments.globals.globals,
+      },
+    },
+    rules: {
+      ...vitest.configs.recommended.rules,
```
agerard-godaddy: Just use the whole config instead of spreading the rules.

You can add it to the top under godaddy configs, i.e.
```diff
export default defineConfig([
  ...goddaddyTypescript,
  ...goddaddyReactTypescript,
+  vitest.configs.recommended,
```

eslint.config.js
```diff
@@ -0,0 +1,50 @@
...
+      'unicorn/filename-case': 'error',
+      'no-sync': 'off',
+
+      /* Gets false positives */
+      'react-hooks/rules-of-hooks': 'off',
```
agerard-godaddy: I'd prefer not to disable this. Instead, let's rename the offending functions.

`useForAllowedPaths -> applyForAllowedPaths` ([here](https://github.com/godaddy/gasket/blob/eslint9/packages/gasket-plugin-middleware/lib/express.js#L45))
and
`useLoader -> isLoader` ([here](https://github.com/godaddy/gasket/blob/eslint9/packages/gasket-plugin-mocha/lib/node-loader-babel/index.js#L85))
seem like more appropriate names anyway.

If we end up with a lot more conflict of react rules in node-code, then we should look into separating the eslint configs.

eslint.config.js
mmason2-godaddy: `no-sync` is deprecated and was turned off in some packages. I'm not sure what the value of that rule is at least in my opinion. I understand the premise but I think given our team and code it's not really necessary. 🤷‍♂️

eslint.config.js
mmason2-godaddy: I can make the changes here 👍

eslint.config.js
mmason2-godaddy: This won't work with plugins.

bbetts-godaddy (approved):

PR #1179, opened 2025-08-20
gasket-cjs examples

mmason2-godaddy:


packages/gasket-cjs/EXAMPLES.md
```diff
@@ -0,0 +1,152 @@
...
+  },
+  swcConfig: {
+    jsc: {
+      target: 'es2018'
+    }
+  }
+});
+```
+
+### `transpileFile`
```
agerard-godaddy: Do no give examples for these other functions, only transpile. These others are not documented in the readme, and are only exported for more specific unit testings. I do not want to gaurantee these APIs will not change

agerard-godaddy (changes requested):

PR #1181, opened 2025-08-20
gasket-data examples

mmason2-godaddy:

### DIFF
--- a/packages/gasket-data/EXAMPLES.md
+++ b/packages/gasket-data/EXAMPLES.md
@@ -1,0 +1,125 @@
+# @gasket/data Examples
+
+This document provides working examples for all exported functions from `@gasket/data`.
+
+## gasketData()
+
+Client-side function to retrieve Gasket data from the DOM. This function looks for a script tag with id `GasketData` and parses its JSON content.
+
+### Basic Usage
+
+```js
+import { gasketData } from '@gasket/data';
+
+// In a React component
+function MyComponent() {
+  const data = gasketData();
+
+  return (
+    <div>
+      <h1>API URL: {data.apiUrl}</h1>
+      <p>Environment: {data.env}</p>
+    </div>
+  );
+}
+```
+
+### In a Next.js Page
+
+```js
+// pages/index.js
+import { gasketData } from '@gasket/data';
+
+export default function HomePage() {
+  const data = gasketData();
+
+  return (
+    <div>
+      <h1>Welcome to {data.appName}</h1>
+      <p>Version: {data.version}</p>
+    </div>
+  );
+}
+```
+
+## resolveGasketData(gasket, req)
+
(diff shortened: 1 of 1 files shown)


packages/gasket-data/EXAMPLES.md
```diff
@@ -0,0 +1,126 @@
...
+      <h1>Welcome to {data.appName}</h1>
+      <p>Version: {data.version}</p>
+    </div>
+  );
+}
+```
+
+## resolveGasketData(gasket, req)
+
+Universal function that works on both client and server. On the client, it calls `gasketData()`. On the server, it calls `gasket.actions.getPublicGasketData(req)`.
```
agerard-godaddy, line 80: You can give an example using getInitialProps - which runs on the server for a first request, then in the browser for client-routing.

packages/gasket-data/EXAMPLES.md
```diff
@@ -0,0 +1,126 @@
+# @gasket/data Examples
+
+This document provides working examples for all exported functions from `@gasket/data`.
+
+## gasketData()
```
agerard-godaddy, line 5: Using this in React components is not a great pattern to replicate. The `useGasketData` React hook is a better fit.

To show this in the browser, perhaps use it in a fetch wrapper and indicate that it must be called in the browser.

```js
import { gasketData } from '@gasket/data';

async function fetchSomeData() {
  if(typeof window !== 'undefined') {
    const data = gasketData();
    return fetch(data.someApiUrl)
  }
}
```

If you want to still show it in a React component, save it for an App Router example and give the example a `'use client'` directive.

packages/gasket-data/EXAMPLES.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+### Server Component (Next.js App Router)
+
+```js
+// app/components/ServerComponent.js
+import { resolveGasketData } from '@gasket/data';
+import gasket from '../gasket.js';
+
+export default async function ServerComponent({ req }) {
```
agerard-godaddy: Server components do not receive a `req` prop.

If you want to show this in a server component, you'll need to use `@gasket/nextjs/request` ([docs](https://github.com/godaddy/gasket/tree/main/packages/gasket-nextjs#request))

packages/gasket-data/EXAMPLES.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+## Common Patterns
+
+### Conditional Rendering Based on Data
+
+```js
+import { gasketData } from '@gasket/data';
+
+function ConditionalFeature() {
```
agerard-godaddy, line 208: Again `gasketData` requires window and will break things for SSR. If you want to keep this example, add `'use client''` to the top.

agerard-godaddy (changes requested):

agerard-godaddy (approved):

PR #1182, opened 2025-08-20
gasket-intl examples

mmason2-godaddy:


packages/gasket-intl/EXAMPLES.md
```diff
@@ -0,0 +1,250 @@
...
+#### locales
+Get the list of supported locales.
+
+```js
+import intlManager from '../path/to/intl.js';
+const supportedLocales = intlManager.locales;
+// ['en-US', 'fr-FR', 'es-ES']
+```
+
+#### defaultLocaleFilePath
```
agerard-godaddy: The `defaultLocaleFilePath` and `staticLocaleFilePaths` properties are really just for internals and do not need examples.

PR #1190, opened 2025-08-21
gasket-plugin-dynamic-plugins examples

mmason2-godaddy:


packages/gasket-plugin-dynamic-plugins/EXAMPLES.md
```diff
@@ -0,0 +1,198 @@
...
+        './test-helpers-plugin.js'
+      ]
+    },
+    staging: {
+      dynamicPlugins: [
+        '@gasket/plugin-docs' // Only docs in staging
+      ]
+    },
+    prod: {
+      dynamicPlugins: [] // No dynamic plugins in production
```
agerard-godaddy: Not a great example - seems to suggest an empty array would stomp over a top level dynamicPlugins. Instead, just have a comment
```suggestion
// No dynamic plugins in production
```

PR #1212, opened 2025-08-22
tests: create-gasket-app, gasket-core, gasket-data, and gasket-intl

jpina1-godaddy:


eslint.config.js
```diff
@@ -100,7 +109,8 @@ export default defineConfig([
     rules: {
       'jsdoc/require-jsdoc': 'off',
       'jsdoc/require-returns': 'off',
-      'jsdoc/require-param': 'off'
+      'jsdoc/require-param': 'off',
+      'jest/no-conditional-expect': 'off'
```
agerard-godaddy: I would like to see this rule kept in place and try to hold to. If there are some exceptions to the rule, they should be allowed on a case-by-case basis. Conditional expects should not be the norm.

packages/create-gasket-app/test/unit/scaffold/actions/generate-files.test.js
```diff
@@ -1,42 +1,105 @@
...
+vi.mock('mkdirp', () => ({ default: mockMkdirpStub }));
+vi.mock('handlebars', () => {
   return {
-    create: () => {
-      const mod = jest.requireActual('handlebars');
-      const handlebars = mod.create();
-      registerHelperSpy = jest.spyOn(handlebars, 'registerHelper');
-      return handlebars;
+    default: {
+      create: () => {
```
agerard-godaddy, line 23: This mock seems very heavy. Can you try using's vitest's equivalent  `vi.importActual`.

You may have to await import outside the mock first. Untested, but something like:

```js
const actualHandlebars = await vi.importActual('handlebars');

jest.mock('handlebars', () => {
  return {
    create: () => {
      const handlebars = actualHandlebars.create();
      registerHelperSpy = jest.spyOn(handlebars, 'registerHelper');
      return handlebars;
    }
  };
});
```

packages/create-gasket-app/test/unit/scaffold/actions/generate-files.test.js
```diff
@@ -1,42 +1,105 @@
...
+        return handlebars;
+      }
     }
   };
 });
-jest.mock('glob', () => {
-  const mod = jest.requireActual('glob');
-  return jest.fn(mod);
-});
+vi.mock('glob', () => ({
```
agerard-godaddy: Same comment as above. Try to use `vi.importActual` to avoid all this mockup code.

packages/gasket-plugin-nextjs/test/next-route.test.js
```diff
@@ -1,75 +1,23 @@
...
-    it('returns valid static route if url has query param', async () => {
-      req.path = '/plans/cohorts/US%20Only';
-
-      const route = await getNextRoute(gasket, req);
-
-      expect(route).not.toBeNull();
-      expect(route.page).toEqual('/plans/cohorts/[cohort]');
-      expect(route.namedRegex.exec(req.path).groups.cohort).toEqual('US%20Only');
-    });
+  it('returns correct route for static path', async () => {
```
agerard-godaddy: Where'd all the other tests go?

agerard-godaddy (changes requested):

agerard-godaddy (changes requested):
A few changes and can you also update the workflow [here](https://github.com/godaddy/gasket/blob/main/.github/workflows/ci.yml#L22
) to use `24.x` which should test against `24.6.x` again.

PR #1215, opened 2025-08-22
gasket-preset-api examples

mmason2-godaddy:


packages/gasket-preset-api/EXAMPLES.md
```diff
@@ -0,0 +1,211 @@
...
+```typescript
+// gasket.ts
+import { makeGasket } from '@gasket/core';
+import pluginExpress from '@gasket/plugin-express';
+import pluginTypescript from '@gasket/plugin-typescript';
+
+export default makeGasket({
+  plugins: [
+    pluginExpress,
+    pluginTypescript,
```
agerard-godaddy: Typescript is a create-only plugin

PR #1217, opened 2025-08-22
gasket-react-intl examples

mmason2-godaddy:

### DIFF
--- a/packages/gasket-react-intl/EXAMPLES.md
+++ b/packages/gasket-react-intl/EXAMPLES.md
@@ -1,0 +1,521 @@
+# @gasket/react-intl Examples
+
+This document provides working examples for all exported functions, components, and hooks from `@gasket/react-intl`.
+
+## Table of Contents
+
+- [LocaleFileStatus](#localefilestatus)
+- [withMessagesProvider](#withmessagesprovider)
+- [useMessages](#usemessages)
+- [useLocaleFile](#uselocalefile)
+- [LocaleFileRequired](#localefilerrequired)
+- [withLocaleFileRequired](#withlocalefilerequired)
+- [ensureArray](#ensurearray)
+- [needsToLoad](#needstoload)
+
+## LocaleFileStatus
+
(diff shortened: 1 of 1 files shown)


packages/gasket-react-intl/EXAMPLES.md
```diff
@@ -0,0 +1,521 @@
...
+```
+
+### Conditional Loading
+
+```javascript
+import { useLocaleFile, LocaleFileStatus, useMessages } from '@gasket/react-intl';
+
+export default function ConditionalComponent({ showAdvanced }) {
+  const basicStatus = useLocaleFile('/locales/basic');
+  const advancedStatus = showAdvanced ? useLocaleFile('/locales/advanced') : LocaleFileStatus.loaded;
```
agerard-godaddy: I don't think you can conditionally use hooks like this. This conditional loading is a bad example. Just delete it's not really useful. Components can conditionally use other components, which in turn could use different locale files.

packages/gasket-react-intl/EXAMPLES.md
```diff
@@ -0,0 +1,521 @@
...
+    <LocalizedInput
+      ref={inputRef}
+      placeholder="Enter text..."
+      onFocus={() => inputRef.current.select()}
+    />
+  );
+}
+```
+
+## ensureArray
```
agerard-godaddy: This is not an exposed function, nor do we want it to be. Delete this section

packages/gasket-react-intl/EXAMPLES.md
```diff
@@ -0,0 +1,521 @@
...
+    </div>
+  );
+}
+
+// Works with single string or array
+<MyComponent localeFiles="/locales/common" />
+<MyComponent localeFiles={['/locales/common', '/locales/forms']} />
+```
+
+## needsToLoad
```
agerard-godaddy: Delete section - This is not an exposed function

agerard-godaddy (changes requested):

PR #1225, opened 2025-09-15
[PFX-978]: Deprecate plugin-service-worker & plugin-workbox

bbetts-godaddy:
## Summary

Ticket: [PFX-978](https://godaddy-corp.atlassian.net/browse/PFX-978)

We will remove the `service-worker` and `workbox` plugins in Gasket 8, so we want to deprecate the usage of it now. 

## Test Plan

- added tests to check if warnings are being logged about deprecated packages


vite-integration-guide.md
```diff
@@ -0,0 +1,404 @@
+# Gasket Vite Integration Guide
```
jpina1-godaddy: Is this vite guide supposed to be here?

bbetts-godaddy: nope! sorry I'll take that out!

jpina1-godaddy (approved):

PR #1235, opened 2025-09-24
Fix broken intl.js imports when locales are in a subdir

mmason2-godaddy:
fix issue where intl.js locale imports are broken when locales are in a subdir

## Summary
- fix issue where intl.js locale imports are broken when locales are in a subdir

## Test Plan
Updated

### DIFF
--- a/packages/gasket-plugin-intl/lib/build-manifest.js
+++ b/packages/gasket-plugin-intl/lib/build-manifest.js
@@ -47,14 +47,12 @@ module.exports = async function buildManifest(gasket, options = {}) {
   const imports = (
     files.map((file) => {
       const managerDir = path.dirname(path.join(tgtRoot, ...managerFilename.split('/')));
-      const relativePath = path.relative(managerDir, tgtLocalesDir);
-      const importName = path.join('.', relativePath, file).replace(/\\/g, '/');
-      console.log('importName', importName);
-      console.log('managerDir', managerDir);
-      console.log('relativePath', relativePath);
+      const relativePath = path.relative(managerDir, tgtLocalesDir) ?? '.';
+      let importName = path.join(relativePath, file).replace(/\\/g, '/');
+      importName = importName.startsWith('.') ? importName : `./${importName}`;
       const keyName = importName
         .replace(/\.json$/, '')
-        .replace('./', '');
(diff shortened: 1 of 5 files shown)


packages/gasket-plugin-intl/lib/build-manifest.js
```diff
@@ -46,7 +46,8 @@ module.exports = async function buildManifest(gasket, options = {}) {
   // generate a content hash for each file
   const imports = (
     files.map((file) => {
-      const importName = ['.', path.basename(localesDir), file].join('/');
+      const relativePath = path.relative(tgtRoot, tgtLocalesDir);
```
jpina1-godaddy: Will you need the `managerDir` path because you need the imports relative to the location of the manager file?

` const managerDir = path.dirname(path.join(tgtRoot, ...managerFilename.split(‘/’)));
 const relativePath = path.relative(managerDir, tgtLocalesDir);
 const importName = path.join(‘.’, relativePath, file).replace(/\\/g, ‘/’);`

mmason2-godaddy: I don't think so.... but maybe I'm not seeing something. I tested this with an Gasket app and it worked fine.

mmason2-godaddy: I see what you mean. Updated.

packages/gasket-plugin-intl/lib/build-manifest.js
```diff
@@ -46,7 +46,12 @@ module.exports = async function buildManifest(gasket, options = {}) {
   // generate a content hash for each file
   const imports = (
     files.map((file) => {
-      const importName = ['.', path.basename(localesDir), file].join('/');
+      const managerDir = path.dirname(path.join(tgtRoot, ...managerFilename.split('/')));
+      const relativePath = path.relative(managerDir, tgtLocalesDir);
+      const importName = path.join('.', relativePath, file).replace(/\\/g, '/');
+      console.log('importName', importName);
+      console.log('managerDir', managerDir);
+      console.log('relativePath', relativePath);
```
jpina1-godaddy: Remove the console logs then ✅

agerard-godaddy (approved):

PR #1426, opened 2026-05-12
Update types to allow actions-only plugins

agerard-godaddy:
## Summary

- Change `Plugin` type so it requires at least one of `hooks` or `actions` (previously `hooks` was always required). Enables action-only plugins.
- Add TypeScript test covering hooks-only, actions-only, both, and the invalid neither case.
- Update plugin-shape wording in docs.

## Test plan

- [x] `packages/gasket-typescript-tests` typecheck passes
- [x] Existing hooks-only plugins still type-check unchanged

### DIFF
--- a/packages/gasket-core/lib/index.d.ts
+++ b/packages/gasket-core/lib/index.d.ts
@@ -142,10 +142,7 @@ export type Hook<Id extends HookId> = HookWithTimings<Id> | HookHandler<Id>;
  *    Plugin Definition   *
  * ----------------------- */
 
-/**
- * A Gasket plugin defines hooks, optional actions, and plugin metadata.
- */
-export type Plugin = {
+type PluginBase = {
   /** Plugin name (required) */
   name: string;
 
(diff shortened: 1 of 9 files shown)


packages/gasket-core/lib/index.d.ts
```diff
@@ -158,20 +155,22 @@ export type Plugin = {
...
 
+type PluginHooks = { [K in HookId]?: Hook<K> };
+type PluginActions = { [K in ActionId]?: ActionHandler<K> };
+
+/**
+ * A Gasket plugin defines hooks and/or actions, plus plugin metadata.
+ * At least one of `hooks` or `actions` must be provided.
+ */
+export type Plugin = PluginBase & (
+  | { hooks: PluginHooks; actions?: PluginActions }
```
jpage-godaddy, line 170: Couldn't you just add the `?` optionality to `hooks` in that `Plugin` type?

jpage-godaddy: Oh I see... you just want to avoid completely empty plugins?

jpage-godaddy (approved):

packages/gasket-core/lib/index.d.ts
agerard-godaddy: I want it to be one or the other, though. A plugin isn't valid unless it declares either `hooks` or `actions`.

jpina1-godaddy (approved):

PR #1439, opened 2026-06-09
feat: upgrade pnpm 10 to 11

mmason2-godaddy:
## Summary

Upgrades pnpm from 10.4.1 to 11.5.2 for the Gasket v8 branch. pnpm 11 brings supply-chain hardening, faster installs (SQLite-backed store), and native publish.

### Changes

- **`.npmrc`** — stripped to registry-only (pnpm 11 only reads auth/registry from `.npmrc`)
- **`pnpm-workspace.yaml`** — added `strictPeerDependencies`, `linkWorkspacePackages`, `preferWorkspacePackages`, `overrides`, and `allowBuilds`
- **`package.json`** — bumped `packageManager` to `pnpm@11.5.2`, removed `pnpm` field, updated `engines.pnpm` to `>=11`
- **`pnpm-lock.yaml`** — regenerated in v11 format
- **Lint fixes** — resolved `vitest/no-unneeded-async-expect-function` errors in `create-gasket-app` and `gasket-plugin-intl`

### Notable pnpm 11 defaults

| Setting | Default | Our override |
|---------|---------|--------------|
| `minimumReleaseAge` | 1 day (blocks fresh publishes) | `0` — can tighten later |
| `strictDepBuilds` | true (blocks build scripts) | Handled via `allowBuilds` map |
| `verifyDepsBeforeRun` | `install` | Kept default |

### Risk

- **Low** — CI workflows need no changes (`pnpm/action-setup@v4` reads `packageManager` automatically)
- **First publish after merge** will validate the native publish flow end-to-end
- No exotic deps, no `.pnpmfile.cjs`, no nested `.npmrc` files

## Test plan

- [x] `pnpm install` completes without errors
- [x] `pnpm run build` succeeds
- [x] `pnpm run test` passes from monorepo root
- [x] `pnpm publish --dry-run` validates native publish flow
- [ ] CI passes on this PR
- [ ] First release from `next` branch publishes successfully

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -2,3 +2,18 @@ packages:
   - packages/*
   - scripts/generate-docs-index
   - tools/template-manager
+
+strictPeerDependencies: true
+linkWorkspacePackages: true
+preferWorkspacePackages: true
+minimumReleaseAge: 0
+
+overrides:
+  react-json-view-lite: "^2.3.0"
+
+allowBuilds:
+  esbuild: true
+  sharp: true
+  storage-engine: true
+  "@scarf/scarf": false
+  core-js: false
--- a/.npmrc
+++ b/.npmrc
@@ -1,5 +1,1 @@
 registry=https://registry.npmjs.org/
-package-manager=pnpm@latest
-strict-peer-dependencies=true
-link-workspace-packages=true
-prefer-workspace-packages=true
--- a/package.json
+++ b/package.json
@@ -92,17 +92,12 @@
     "vitest": "^3.2.0"
   },
   "engines": {
-    "pnpm": ">=8",
+    "pnpm": ">=11",
     "npm": "please-use-pnpm",
     "yarn": "please-use-pnpm",
     "node": ">=24"
   },
-  "packageManager": "pnpm@10.4.1",
-  "pnpm": {
-    "overrides": {
-      "react-json-view-lite": "^2.3.0"
-    }
-  },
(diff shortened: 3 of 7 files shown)


agerard-godaddy (approved):

jpina1-godaddy (approved):

pnpm-workspace.yaml
```diff
@@ -2,3 +2,18 @@ packages:
   - packages/*
   - scripts/generate-docs-index
   - tools/template-manager
+
+strictPeerDependencies: true
+linkWorkspacePackages: true
+preferWorkspacePackages: true
+minimumReleaseAge: 0
```
kbader-godaddy, line 9: I might be missing context here, but is `minimumReleaseAge: 0` meant to be permanent? My understanding is pnpm 11 defaults it to 1440 (24h) to keep freshly-published versions from resolving before they can be flagged, and since our own packages are workspace-linked rather than registry-resolved, I wasn't sure the canary/next publish flow actually needs it disabled. Curious whether dropping the line (to inherit the default) would still work, or if there's a publish case I'm overlooking?

kbader-godaddy (approved):

pnpm-workspace.yaml
mmason2-godaddy: You're right about workspace packages, but the issue is transitive deps. The default 24h policy rejects lockfile entries for any package published within that window, which causes pnpm install to fail. Needed to keep it at 0 to avoid sporadic CI failures.

PR #1440, opened 2026-06-09
feat: fallow CI gate

kawikabader:
> **TL;DR:** Adds the Fallow CI gate (blocks PRs that introduce dead code, unused deps, duplication, complexity) plus a sticky health-score comment. Fixes `gasket-plugin-https` as a worked example. First of several PRs; follow-ups drain the remaining repo-wide findings.

## Summary

Pull requests now run a [Fallow](https://github.com/fallow-rs/fallow) gate that catches dead code, unused dependencies, duplication, and complexity creep before they land. The gate is changeset-scoped: it fails only on issues a PR *introduces*. The ~340 pre-existing findings (mostly plugin-architecture false positives from dynamic loading, generator templates, and 316 plugin entry points) register as inherited and never block a PR until someone edits the file that carries them. No baseline file to maintain — the codebase drains gradually as edit traffic flows through it.

Every PR also gets a sticky health-score comment, so the repo-wide grade is visible and trends upward as touched files get cleaned. The `gasket-plugin-https` changes here are a worked example of that drain-on-edit model: a real package taken to zero findings while passing through the gate.

## How the gate works

`fallow audit` runs new-only attribution against the PR base: it diffs the changeset, marks each finding introduced or inherited, and gates only on introduced error-severity findings. Verified against `next` — a PR reports `dead_code_introduced: 0, dead_code_inherited: 151`, verdict `pass`.

Three non-obvious things the workflow gets right:

- **`fetch-depth: 0`** — the default shallow clone lacks the base ref, so `--base` can't resolve.
- **`origin/`-prefixed base** — `github.base_ref` is the bare name `next`, which doesn't resolve in a CI checkout; only `origin/next` exists. Passed via `env:` so a branch name can never be read as shell.
- **Exit code is the gate** — `fallow audit` exits 0 (clean / warns only), 1 (introduced error-severity → fail), or 2 (runtime error → fail loudly). No JSON parsing.

## Health-score comment

A second, non-gating step posts one sticky comment per PR, updated in place on every push. It renders the letter grade with a status emoji and a penalty breakdown table — non-zero penalties only, largest first — so reviewers see what's costing points repo-wide and watch the grade climb as cleanup lands.

- `if: always()` + `|| true` — reports even when the gate fails; a health hiccup never flips the job outcome.
- Sticky via a hidden `<!-- fallow-health-score -->` marker: find-and-PATCH the existing comment, else POST a new one.
- No-ops safely on fork PRs (read-only `GITHUB_TOKEN`); uses `pull_request`, never `pull_request_target`.

## False positives

Suppressed at the source, closest to the cause:

| Case | Mechanism |
|---|---|
| One-off (dynamic require, intentional unused export) | inline `// fallow-ignore-next-line <issue-type>` |
| Whole directory (`generator/`) | `ignorePatterns` in `.fallowrc.json` |
| Reference-only dep (`@gasket/plugin-metadata`) | `ignoreDependencies` in `.fallowrc.json` |
| Recurring shape | `.fallowrc.json` over scattered inline comments |

CONTRIBUTING.md documents the decision tree plus a manual `fallow fix` cleanup loop for draining existing debt on demand.

## gasket-plugin-https cleanup

The first package run through the gate, as a reference for the drain-on-edit pattern:

- Dead dependency dropped; `@gasket/plugin-metadata` false positive ignored at source.
- Server-logging helpers extracted; lint warnings (loose `Function`, bare statements) cleared.
- Internal types moved to `internal.d.ts`; `terminus` opts and `startServer` typed concretely against the public `ActionHandler`.
- Tests parameterized and branches covered to 100%, with an 80% coverage floor added to `vitest.config.js` (`lib/**/*.js` only — `index.d.ts` has no runtime to exercise).

## Notes

- `fallow` added as a root dev-dependency (`^2.91.0`, caret per repo convention).
- Unused deps removed across `create-gasket-app`, `gasket-plugin-docs`, `gasket-plugin-swagger`, `gasket-react-intl`, and others.
- Known tradeoff: `pnpm install`'s postinstall runs a full monorepo build on each gate run (same baseline as `ci.yml`). Acceptable for now; revisitable with `--ignore-scripts` if fallow proves it doesn't need build artifacts.

## Test plan

Validate by opening throwaway PRs against the workflow:

- PR adding an unused export in a touched file → gate fails, finding in the log.
- PR touching only clean files → passes despite the ~340 repo-wide findings.
- PR with a broken base ref → fails loudly (exit 2), not a silent pass.
- Health comment appears once per PR and updates in place across pushes.


packages/gasket-typescript-tests/test/plugin-https.spec.ts
```diff
@@ -122,7 +121,6 @@ describe('@gasket/plugin-https', () => {
 
   it('adds the beforeShutdown lifecycle', () => {
     const handler: Hook<'beforeShutdown'> = async (gasket) => {
-      // await shutdownTheThing();
```
agerard-godaddy, line 125: Did this example code comment get caught because it looks like code?

Instead of leaving this with an empty body, could we drop a simple non-code sentence comment?

Something like:
```js
// perform shutdown logic
```

packages/gasket-typescript-tests/test/plugin-https.spec.ts
```diff
@@ -134,7 +132,6 @@ describe('@gasket/plugin-https', () => {
 
   it('adds the onShutdown lifecycle', () => {
     const handler: Hook<'onShutdown'> = async (gasket) => {
-      // await cleanupStuff();
```
agerard-godaddy, line 137: Same a previous comment

.github/workflows/fallow.yml
```diff
@@ -0,0 +1,126 @@
+name: Fallow
+
+on:
+  workflow_dispatch:
+  pull_request:
```
agerard-godaddy, line 5: Let's limit fallow to the `next` branch to limit churn on `main` right now

```suggestion
  pull_request:
    branches:
    - next
```

.github/workflows/fallow.yml
```diff
@@ -0,0 +1,126 @@
...
+                "",
+                "What is costing points, repo-wide:",
+                "",
+                table,
+                "",
+                `<sub>Repo-wide score for trend visibility — not a merge gate. ` +
+                `Watch the grade climb as PRs clean up touched files. Updated each push (\`${sha}\`).</sub>`
+              ].join("\n");
+            }
+            fs.writeFileSync("fallow-comment.md", md + "\n");
```
agerard-godaddy: To be safe, gitignore this temp `fallow-comment.md` file, or perhaps write it to the `.fallow/` directory which you've already gitignored.

packages/gasket-typescript-tests/test/plugin-https.spec.ts
kbader-godaddy: Yeah, fallow read the original `// await shutdownTheThing();` as commented-out code. I wasn't totally sure what the intent of that placeholder was, so I just removed it, which left the body empty. Added `// perform shutdown logic` per your suggestion. Thanks!

packages/gasket-typescript-tests/test/plugin-https.spec.ts
kbader-godaddy: Same fix here, dropped the old `// await cleanupStuff();` placeholder and added `// perform shutdown logic`. Thanks!

.github/workflows/fallow.yml
kbader-godaddy: Done, scoped `pull_request` to the `next` branch. Thanks!

.github/workflows/fallow.yml
kbader-godaddy: Good catch. Moved both temp files (`health.json` and the comment) into the `.fallow/` dir. Thanks!

agerard-godaddy (approved):

packages/gasket-plugin-https/lib/actions.js
```diff
@@ -7,11 +7,96 @@ import { getPortFallback, portInUseError, startProxy, getRawServerConfig } from
...
 
+/**
+ * Log the create-servers failure, distinguishing port-in-use from other causes.
+ * @type {import('./internal.d.ts').logServerError}
+ */
+function logServerError(errors, serverOpts, logger) {
+  const message = portInUseError(errors)
+    ? 'Port is already in use. Please ensure you are not running the same process from another terminal!'
+    : `Failed to start the web servers: ${errors.message}`;
+  const errorMessage = errs.create({ message, serverOpts });
```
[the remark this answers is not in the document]
kbader-godaddy: In create-servers v3.3.0 the top-level callback never receives a bare array; `createServers` wraps failures via `errs.create({ message: errorSource[0].message, ... })`, normalizing the array itself before the callback runs, so `errors.message` is always set here (never undefined). The `Array.isArray` guard in `portInUseError` is belt-and-suspenders for an older contract. Happy to mirror that normalization in `logServerError` for type consistency if you'd prefer, but it isn't reachable in practice.

